PR #2, opened 2021-08-05
SidebarLayout: if not item.openInTab, target='_self'

DavidLarsKetch:
The sidebar was opening links in a new tab when it should not have been. Setting `<a target="_self">` instead of `<a target=false>` resolves that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@xy-planning-network/trees",
-  "version": "0.2.53",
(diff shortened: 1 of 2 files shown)


profsmallpine:
Nice fix, you need to manually bump the version in package.json (CI will send this over to NPM so you can import the update).

profsmallpine (commented):
Hot toddy

PR #4, opened 2021-08-23
refactor: content modal

steven-klein:
# What this does

- updates the content modal component dropping the required content prop in favor of using it's default slot
- updates docs accordingly

### DIFF
--- a/dev/content/Overlays.vue
+++ b/dev/content/Overlays.vue
@@ -3,7 +3,8 @@
     <div class="max-w-3xl mx-auto">
       <ComponentLayout title="Content Modal">
         <template v-slot:description>
-          This bad boy is used to show the user some content.
+          This bad boy is used to show the user some content. The default slot
+          is for the primary content.
         </template>
 
         <div>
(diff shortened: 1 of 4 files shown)


DavidLarsKetch (commented):
@steven-klein I noticed the dirty `package-lock.json` as well in my last set of changes and chose to let sleeping dogs lie. What's your take on it? I thought `"dev": true` was always attached to packages in `devDependecies`, so surprised to see it stripped in many places, but, outside my wheelhouse here.

DavidLarsKetch (approved):

steven-klein:
It's hard to say why exactly those dev:true properties are being dropped, but as far as I can tell it's related to the update on @headlessui-vue.  Somewhere in the chain of dependency a package must using these as a non-dev dependency.  In our case I think it will be okay as nothing should change as far as what is published to npm.  We may need to think through whether dependencies like @headlessui-vue should be "pinned" to avoid introducing a change we're not testing for, but looking through their changelog it looks like a safe update.

PR #5, opened 2021-09-17
proposal: form field expansion

steven-klein:
# What this does.

- Expands on the current set of input fields.
- Includes/exposes a `<BaseInput>` component for the most common text like inputs.
- includes/exposes an `<InputLabel>` component for custom use.
- Includes/exposes an `<InputHelp>` component for custom use.
- Simplifies adding labels and help text to most input types that can support them via optional props.
- Removes the redundancy of using props for common input attributes like required and disabled by using $attrs.
- Adds basic A11y support to all current input types with no consumer configuration by using a dynamic UUID.
- Should maintain compatibility with current usage.

This is part 1 of a multi-phase plan for expanding on form fields and form handling in general.  Primary goal in this round was to expand on the current fields to make them easier to use, particularly with labels and help text, as well as, adding some zero config accessibility features and reducing the need for composable styles written in css outside of the templates.

Future rounds will focus on the form, fieldset, and legend components and their A11y concerns, as well as for validation and accessible error display for inputs.

## TODO:

- [x] add a textarea component consistent with the other fields
- [x] test changes on college try
- [ ] test changes on archive
- [x] update the docs, specifically props and the use of attributes like required and disabled.  Left as is to verify compatibility with current usage.
- [x] need to import InputLabel and InputHelper into the field components that use them as components for situations where the consumer only includes select components from trees

## NPM Link notes

When testing on college try or archive and using npm link on trees you may need to delete the node_modules directory in trees to avoid resolving two different instances of vue.

### DIFF
--- a/src/lib-components/forms/Checkbox.vue
+++ b/src/lib-components/forms/Checkbox.vue
@@ -19,8 +19,7 @@
       <label
         :id="`${uuid}-label`"
         :for="uuid"
-        class="text-gray-500"
-        :class="{ 'font-medium': emphasis }"
+        :class="{ 'font-semibold': emphasis }"
         v-text="label"
       ></label>
     </div>
--- a/src/lib-components/forms/Checkbox.vue
+++ b/src/lib-components/forms/Checkbox.vue
@@ -2,16 +2,23 @@
   <div class="relative flex items-start">
     <div class="h-5 flex items-center">
       <input
-        type="checkbox"
-        class="focus:ring-blue-500 h-4 w-4 text-xy-blue border-gray-300 rounded disabled:opacity-50 disabled:cursor-not-allowed"
+        :aria-labelledby="label ? `${uuid}-label` : undefined"
         :checked="modelValue"
-        :disabled="disabled"
-        @change="$emit('update:modelValue', $event.target.checked)"
-        :required="required"
+        class="focus:ring-blue-500 h-4 w-4 text-xy-blue border-gray-300 rounded disabled:opacity-50 disabled:cursor-not-allowed"
+        :id="uuid"
+        type="checkbox"
+        v-bind="{
+          ...$attrs,
+          onChange: ($event) => {
+            $emit('update:modelValue', $event.target.checked);
+          },
+        }"
       />
     </div>
     <div class="ml-3 text-sm leading-5">
       <label
+        :id="`${uuid}-label`"
+        :for="uuid"
         class="text-gray-500"
         :class="{ 'font-medium': emphasis }"
         v-text="label"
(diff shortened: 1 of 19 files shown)


src/lib-components/forms/Checkbox.vue
```diff
@@ -2,16 +2,23 @@
...
+        :aria-labelledby="label ? `${uuid}-label` : undefined"
         :checked="modelValue"
-        :disabled="disabled"
-        @change="$emit('update:modelValue', $event.target.checked)"
-        :required="required"
+        class="focus:ring-blue-500 h-4 w-4 text-xy-blue border-gray-300 rounded disabled:opacity-50 disabled:cursor-not-allowed"
+        :id="uuid"
+        type="checkbox"
+        v-bind="{
+          ...$attrs,
```
profsmallpine, line 11: 🤑

src/lib-components/forms/DateRangePicker.vue
```diff
@@ -1,23 +1,33 @@
...
-    class="shadow-sm focus:ring-blue-500 focus:border-blue-500 block w-full sm:text-sm border-gray-300 rounded-md"
-    placeholder="mm-dd-yyyy range"
-  />
+  <InputLabel
+    class="block"
+    :id="`${uuid}-label`"
+    :for="uuid"
+    :label="label"
+  ></InputLabel>
+  <BaseInput type="text" placeholder="mm-dd-yyyy range" :id="uuid"></BaseInput>
```
profsmallpine, line 8: My mind is getting a little jelly after the day, but 1. where is `BaseInput` coming from and 2. doesn't the base includ label and help?

profsmallpine (approved):
The use of $attrs is a nice win for componetizing the inputs. I look forward to seeing this concept mature

src/lib-components/forms/Checkbox.vue
steven-klein: definitely comes with it's own set of concerns, but has a lot of potential

src/lib-components/forms/DateRangePicker.vue
steven-klein: Good find.  Must have overlooked importing `BaseInput` when debugging with the dev site, since it makes use of `app.use(Trees);` and it became globally available.  It should be imported now and simplified to just using `BaseInput` with its label and help props.  Initially, it was composed with the label and help components to make initializing flatpickr easier with the uuid, but this reminded me that really the uuid should check for an $attrs.id before creating a unique.  So `uuid = (this.$attrs.id as string) || Uniques.CreateIdAttribute();` is being used which should play nicely with custom ID attributes that continue to work with aria tags.

PR #13, opened 2022-02-07
Adds <Search /> component

DavidLarsKetch:
- `<Search />` is a way of styling an `<input />` as a search bar and limiting the number of updates by debouncing the `input` event
  - Replaces watching `change` event as `<Table />` did previously, pushing towards more ergonomic behavior of the component
  - I did not use `<BaseInput />` due to the lack of flexibility adding in the `<SearchIcon />`, it may very well be abstracting that component further is a bridge too far and the duplicated code is acceptable for now
- Replaces existing markup in `<Table />` with new `<Search />`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,8 @@
         "@headlessui/vue": "^1.4.2",
         "@heroicons/vue": "^1.0.5",
         "axios": "^0.21.4",
-        "flatpickr": "^4.6.9"
+        "flatpickr": "^4.6.9",
+        "lodash.debounce": "^4.0.8"
       },
       "devDependencies": {
         "@types/node": "^16.11.13",
@@ -3293,6 +3294,11 @@
       "version": "4.17.21",
       "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.21.tgz",
       "integrity": "sha512-v2kDEe57lecTulaDIuNTPy3Ry4gLGJ6Z1O3vE1krgXZNrsQ+LFTGHVxVjcXPs17LhbZVGedAJv8XZ1tvj5FvSg=="
+    },
+    "node_modules/lodash.debounce": {
+      "version": "4.0.8",
+      "resolved": "https://registry.npmjs.org/lodash.debounce/-/lodash.debounce-4.0.8.tgz",
+      "integrity": "sha1-gteb/zCmfEAF/9XiUVMArZyk168="
     },
     "node_modules/lodash.merge": {
       "version": "4.6.2",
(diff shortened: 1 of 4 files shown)


steven-klein:
@DavidLarsKetch @profsmallpine  

I know this is already closed as a result of a quick offline convo I had with dlk, but I wanted to drop some thoughts in here in case we ever need to return to them.

After thinking about it, I'm not sure the functionality of debouncing should be tied to a component.  There's limited reusability to the pattern and it tightly scopes functionality to something that is primarily concerned with simple UI presentation.  A directive or input binding modifier could have been an interesting way to have a reusable way to debounce input fields of any type, but there's always the possibility that we want to know the updated model value even though some other event should be debounced, so decoupling it seems to be the best overall option.

A utility function made available by Trees could possibly be as simple as this:

```ts
function debounce(func: () => void, timeout = 500) {
  let timer: NodeJS.Timer
  return (...args: any[]) => {
    clearTimeout(timer)
    timer = setTimeout(() => {
      func.apply(this, args)
    }, timeout)
  }
}
```

I do think there might be some merits to the lodash implementation which has advanced configuration options and additional methods like flush and cancel.  `import { debounce } from "lodash-es"` would probably be the way to go for an individual application to limit its footprint.  Additionally there is some really interesting stuff happening with [VueUse](https://vueuse.org/shared/usedebouncefn/#usedebouncefn) that we should keep an eye on.

Probably the most compelling reason not to go down this road though, is that it only solves half the problem which is, debouncing is great for limiting the total number of api calls that go out, but it doesn't prevent a previous request from resolving later than a subsequent request.  This isn't likely to be a huge issue for us currently, but it's possibly the most important regarding what we're discussing.

I don't have a great pattern for cancelling network requests, but Axios cancel tokens should offer it.  It's not something I think we need to prioritize, but it's on my mind for a more complete solution.

In terms of just visually seeing how we might implement just a straight debounce right now here's a contrived example with lodash-es.

```vue
<script setup lang="ts">
import { debounce } from "lodash-es"
import { ref } from "vue"

const searchQuery = ref("")
const apiData = ref(0)

const myApiCall = () => {
  console.log("debounced fake api call...")

  new Promise((resolve) => {
    // timeout to emulate network request
    setTimeout(() => {
      resolve(1)
    }, 2000)
  }).then((inc) => {
    console.log("api data updated...")
    apiData.value += inc
  })
}

const updateSearch = debounce(myApiCall, 500)
</script>

<template>
  <BaseInput
    v-model="searchQuery"
    type="search"
    label="Search"
    @input="updateSearch"
  />

  <div>Query: {{ searchQuery }}</div>

  <div>API Data: {{ apiData }}</div>
</template>
```

PR #14, opened 2022-02-08
feat: Popover Component

steven-klein:
# What this does

- Implements a bare bones version of the headless ui popover.
- Provides convenience props for trigger text/icon for most commonly expected trigger use cases.
- Provides convenience text prop for simple tooltip like popovers.
- Adds positioning prop for flexible positioning.
- Keeps presentation styles minimal leaving it mostly up to the PositionContent slot to do whatever it needs.

@profsmallpine something for us to consider as we add features like this is how opinionated and convenient they should be.  Example we could strip out the triggerText and triggerIcon props and instead use those in more highly opinionated components such as "Tooltip" where the styling would be more opinionated for that specific use case.

https://user-images.githubusercontent.com/3856703/153058513-bcaf9947-486d-4e74-80f0-ced0158abc6d.mov

# TODO

- [x] bump version number

### DIFF
--- a/dev/content/Overlays.vue
+++ b/dev/content/Overlays.vue
@@ -1,6 +1,8 @@
 <script setup lang="ts">
 import { ref } from "vue"
 import { CheckIcon } from "@heroicons/vue/outline"
+import { ExclamationIcon } from "@heroicons/vue/outline"
+import { PopoverPosition } from "@/lib-components/overlays/Popover/Popover.vue"
 
 const contentModalCopy = `<ContentModal v-model="open" :content="content" :title="title"></ContentModal>`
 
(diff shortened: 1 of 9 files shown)


profsmallpine:
Code looks good, appreciate the ease of overriding classes in the content slot. I'd opt to start with as many opinions as possible (in general). Styling can be what we want for the existing use case (expanded in the future if need be). That way, if someone is being lazy, they will just get the default UI which will help us maintain features working the same way across apps and helping solidify user expectations of when something is clickable, etc..

In that vein, I'd consider auto-positioning (depending on how sane/reliable it is) so that you just drop the content in and are off n running. Stripping away the trigger text/icon makes a lot of sense here (moving towards a tooltip, maybe built off of this). I'm curious if we could just make it a convention to use the little i icon as it becomes muscle memory to see that and know more info lies beneath (would we ever use the popover or is a more opinionated tooltip enough). Those types of decisions help us tighten up the UI, even if it's less flexible in the end. Open for more discussion though if you have different intuitions.

dev/content/Overlays.vue
```diff
@@ -193,6 +224,104 @@ const spinner = function (): void {
...
+            </Popover>
+          </div>
+
+          <div class="mt-2 flex justify-center">
+            <Popover>
+              <PopoverContent
+                class="bg-xy-blue text-white border-0 w-screen lg:max-w-md"
+              >
+                <div class="text-base p-4">
+                  Hi, hello, nice to meet you in the flesh I've only seen you
```
profsmallpine, line 293: Beautiful 😢

profsmallpine (approved):
👍

PR #18, opened 2022-05-17
feat: use base api

steven-klein:
# What this does

- adds `withDelay` and `dataIntercept` options to the `RequestOptions` interface.  `withDelay` artificially delays the response from the API request for UI purposes.  `dataIntercept` is a convenience option to receive the response result that is nested inside of AxiosResponse data and server data fields
- adds type generics to BaseAPI methods so that the success response can be correctly inferred
- adds composable `useBaseAPI` which returns a handful of reactive variables that are generally useful for managing the state of a component
- adds a `debounceLeading` function for button triggered debouncing where trailing clicks should be debounced but not the initial click
- adds a composables page to the docs

## Note:

- the abort feature in useBaseAPI requires an axios version bump
- https://github.com/xy-planning-network/college-try/pull/342 is currently using useBaseAPI via the dev release on npm

### DIFF
--- a/dev/content/Composables.vue
+++ b/dev/content/Composables.vue
@@ -5,24 +5,27 @@ import { computed } from "vue"
 import useBaseAPI from "../../src/composables/useBaseAPI"
 import { debounceLeading } from "../../src/helpers/Debounce"
 
-interface Thing {
-  id: number 
-  created_at: number
-  type: string
-  title: string
-}
-
-interface ThingResponse {
+interface Paginated<T> {
   page: number
   perPage: number
   totalItems: number
   totalPages: number
-  items: Thing[]
+  items: T[]
+}
+
+interface Conifer {
+  id: number 
+  name: number
+  type: string
+  leaf: {
+    type: string
+    length: string
+  }
 }
 
 const { result, error, isLoading, isFinished, isAborted, hasFetched, execute, abort } =
-  useBaseAPI<ThingResponse>(
-    "https://my-json-server.typicode.com/xy-planning-network/trees/things",
+  useBaseAPI<Paginated<Conifer>>(
+    "https://my-json-server.typicode.com/xy-planning-network/trees/conifers",
     "GET",
     { dataIntercept: true, withCredentials: false}
   )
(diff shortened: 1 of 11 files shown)


steven-klein:
@DavidLarsKetch @profsmallpine it's not always immediately obvious why you might what to use `useBaseAPI`.  But I think it sets the table for making richer UI's easier, particularly loading screens.  It will be particularly useful in list-like components  where we want to bake in some API handling.

Additionally, the dataIntercept option works hand in hand with the additional of the Type generics on the api methods.  It's all back compatible by defaulting to any.  The goal here was to start moving toward our 99% use cases where an API call returns an AxiosResponse which has a data key, but the Trails response also has another data key.  Most of the time we want what's on the Trails response data key and that is what dataIntercept resolves.  If you add your expected response Type to the generic on your call, then type inference should already know what your `result/success` variable is.

profsmallpine:
@steven-klein what was the impetus (feature) that kicked this off?

steven-klein:
> @steven-klein what was the impetus (feature) that kicked this off?

There wasn't necessarily a singular feature that really kicked it off, it was primarily just recognizing how much boilerplate does or could go exist for any piece of interface that needs to interact with the API and wanting to make it easier to handle states like loading, disabled, error etc. The tax status list is a good example of something that could have benefited from something like this.  Or any current and future list like components should be able to make good use of it.  

The composition function is likely more useful for constructing new opinionated trees components (dynamic lists/tables, forms etc.) or one off complex app specific components like the tax status page.  The additions to the BaseAPI is a mixed bag, but was done mostly for convenience.  The delay feature makes it trivial to leverage the existing full screen loader for nice feedback and the data intercept reduces Type declaration complexities.

dev/content/Composables.vue
```diff
@@ -0,0 +1,187 @@
+<script setup lang="ts">
+import { RequestOptions } from "@/api/base"
+import { AxiosError } from "axios";
+import { computed } from "vue"
+import useBaseAPI from "../../src/composables/useBaseAPI"
+import { debounceLeading } from "../../src/helpers/Debounce"
+
+interface Thing {
```
DavidLarsKetch, line 8: Motion to change this to some coniferous tree related analogy, like `Fir` or `Spruce`.

PR #20, opened 2022-06-16
Feat: flashes and spinner composables

steven-klein:
# What this does

- adds a "persistent" argument to flash methods allowing them to be displayed as "sticky" (must be removed by the user).
- exports the xy-spinner as an isolated component for alternative indicator usage
- migrates flashes and spinner from events to composables
- exposes a global (app) composable for flashes and spinner `useAppFlashes` and `useAppSpinner`
- adds convenience methods for flash types `error, warning, info, success` alongside `genericError`
- adds configuration for default error email and removeDelay on flashes
- improves typescript integration
- maintains backward compatibility with current flash and spinner events on the VueBus

## Why

There's nothing technically wrong with using Mitt and a global event bus.  It generally gets the job done, but establishing your listeners at the right time so that you can do something useful with it can be difficult.  A primary example is that the Flash and Spinner components are mounted late, sometimes requiring the use of NextTick in child components. Using NextTick is a fine solution, but not an intuitive one.  

There are a number of alternative solutions to that specific problem, but they don't necessarily resolve the ergonomics of the current VueBus usage:

`Window.VueBus.emit("a/complex/string", anUntypedArgument)`

It's a minor inconvenience and can likely can be resolved with stronger type definitions on VueBus.  But why rely on both the Window object and an additional dependency throughout our components?  The composition API gives us a reusable reactive pattern that could support similar components (ex: toasts or other flash like notifications, component specific indicators), simplifies queuing objects and/or toggling state, and makes it shareable in a familiar ES6 module pattern.

## Can't we similarly wrap up events into convenient functionality like a composable?

Yes, I started there.  But the composition API is familiar in that you just write JavaScript functions that happen to include reactivity, and anything can choose to react to those changes.

Using the composition API isn't an enormous reduction in complexity -- it's a new type of complexity, but a familiar one, it follows Vue convention, gets easier to understand with each new use case, and has decent type support ready to go.

## What's with the function call in `useAppFlashes().info("hi")` and `useAppSpinner().show()`?

This is primarily out of convention.  Composables in Vue are functions that return something that includes reactivity.  Technically, these are wrapping up a shared instance of a composable.   The use of the "use" prefix may be enough of an indicator to understand you're working with reactivity.

Additionally, I didn't want to initialize these global instances until required/used.  Presumably though, with an ES6 module if you've imported it, you plan to use it and only then will it be in the bundle, so why not just instantiate it and call it a day.  They are also pretty light weight.  If everyone would rather simply `useAppFlashes.info("hi")` I can refactor.


steven-klein:
API
```ts
import {useAppFlasher, useAppFlashes, useAppSpinner} from "@xy-planning-network/trees"

// basics
useAppFlashes().error("error")
useAppFlashes().info("info")
useAppFlashes().success("success")
useAppFlashes().warning("warning")

// generic errors
//
// Will show a default message without an email address, unless one has been configured on the instance with useAppFlasher().configure({email: "support@xyplanningnetwork.com"})
useAppFlashes().genericError()
// override the configured email
useAppFlashes().genericError("help@xyplanningnetwork.com")

// persistent
useAppFlashes().warning("warning", true)

// spinner
useAppSpinner().show()
useAppSpinner().hide()

```

dev/content/Overlays.vue
```diff
@@ -128,18 +125,74 @@ const tooltipCopy = `<Tooltip>Here's something subtly helpful.</Tooltip>`
...
-            <button type="button" class="xy-btn" @click="flash()">
-              Show Me
-            </button>
-          </div>
+          <ul class="mt-1 space-y-2">
+            <li>
+              <InputLabel
+                label="Flash generic error using configured email address:"
+              />
+              <pre><code class="language-typescript">/* Import and do this anytime, best before createApp() or inside a root app level component */
```
profsmallpine, line 137: This UI is not breaking in its container but overflowing out the edges

src/lib-components/overlays/Flash.vue
```diff
@@ -1,80 +1,40 @@
...
-    type: "error",
-    message:
-      "Whoops! Something went wrong, please reach out to " +
-      `<a class="underline text-xy-blue" href="mailto:${email}">${email}</a>` +
-      " if the issue persists.",
-  })
-}
-
 onMounted(() => {
+  // NOTE: (spk) here for backwards compatibility
```
profsmallpine, line 24: I'm down to remove this here, and migrate archive/tax/portal rather than making it backwards compatible. It's not that large of a task, and it helps us move to a world without mitt.

profsmallpine (changes requested):
Biggest proposed change is removing backwards compat.

dev/content/Overlays.vue
steven-klein: Ah, prism.js wasn't kicking in on the dynamic page change.  Made a quick update so that it works as expected, but will plan to revisit code samples in general.

src/lib-components/overlays/Flash.vue
steven-klein: That sounds good to me as well.  I wasn't sure of everyone's appetite for breaking changes, but agreed this should be a pretty quick and easy update.

PR #21, opened 2022-06-16
feat: radio cards

steven-klein:
# What this does

- adds a fancy radio button that looks like a card
- support for HTML5 form validation and FormData
- includes the benefits of headless ui features like focus trap and aria tags
- avoids using the headless ui feature of supporting FormData since their solution does not support the required attribute
- fixes layout bug in radio buttons in columns mode
- brings radio/checkbox legend styles inline with field level labels

### Why

- a potential use case in Portal.
- seems generally beneficial to have additional form field display options when the classics aren't enough

https://user-images.githubusercontent.com/3856703/174171772-503bf766-1021-4ae5-b5eb-589c535bdcc5.mov

### DIFF
--- a/src/lib-components/forms/RadioCards.vue
+++ b/src/lib-components/forms/RadioCards.vue
@@ -131,12 +131,21 @@ const nameAttr = computed(() => {
               <RadioGroupDescription v-if="option.help" as="div">
                 <InputHelp tag="div" class="mt-auto" :text="option.help" />
               </RadioGroupDescription>
-              <div v-if="option.sublabel" class="mt-auto mb-0">
-                <RadioGroupDescription as="div" class="mt-4">
-                  <InputLabel
-                    tag="span"
-                    class="mt-auto mb-auto"
-                    :label="option.sublabel"
+              <div
+                v-if="option.sublabel || $slots.sublabel"
+                class="mt-auto mb-0"
+              >
+                <RadioGroupDescription
+                  as="div"
+                  class="font-semibold leading-snug mt-4 text-gray-900 text-sm"
+                >
+                  {{ option.sublabel }}
+                  <slot
+                    name="sublabel"
+                    :active="active"
+                    :checked="checked"
+                    :disabled="disabled"
+                    :option="option"
                   />
                 </RadioGroupDescription>
               </div>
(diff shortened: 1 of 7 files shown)


src/lib-components/forms/RadioCards.vue
```diff
@@ -0,0 +1,183 @@
...
+    modelValue: undefined,
+  }
+)
+
+const emit = defineEmits<{
+  (e: "update:modelValue", modelValue: ModelValue): void
+}>()
+
+const attrs = useAttrs()
+const uuid = (attrs.id as string) || Uniques.CreateIdAttribute()
```
profsmallpine, line 55: any reason to always use: Uniques.CreateIdAttribute()? IME, it's not uncommon to copy pasta the same id by accident in a form

src/lib-components/forms/RadioCards.vue
```diff
@@ -0,0 +1,183 @@
...
+const nameAttr = computed(() => {
+  return typeof attrs.name === "string" && attrs.name !== "" ? attrs.name : uuid
+})
+</script>
+
+<template>
+  <RadioGroup
+    :modelValue="checked"
+    @update:model-value="onChange"
+    :disabled="typeof attrs.disabled === 'boolean' ? attrs.disabled : false"
```
profsmallpine, line 87: Nice little ditty here

src/lib-components/forms/RadioCards.vue
```diff
@@ -0,0 +1,183 @@
...
+    :aria-invalid="invalid === true ? 'true' : null"
+    :aria-errormessage="invalid === true ? `error-${uuid}` : null"
+  >
+    <RadioGroupLabel v-if="legend" class="block">
```
profsmallpine, line 91: Assuming this markup primarily came from headless/tailwindui so not diving too deep here

steven-klein: Yeah, aside from componetizing and html5 required validation support it's very close to a copy paste of tailwind UI wrapping headless.

src/lib-components/forms/RadioCards.vue
steven-klein: Good point, we had some cases where it made sense on basic form components, but we never merge attributes here, so it's not as useful.  Let's use `Uniques` until we find a reason to allow a custom ID attribute.

PR #24, opened 2022-12-14
refactor/fix: overlay stacking context and z-index values

steven-klein:
# What this does

- Resolves the issue where components like flash and spinner are visibly behind other overlay components such as slideover and modal by wrapping the spinner and flash components in the headlessui Portal component so that all overlay full viewport coverage components share the same staking context.
- Updates existing z-index values on overlay components according to each component's perceived display importance and document those new values.
- Adds a minor drop-shadow on the spinner component to increase visibility in the case of multiple full viewport overlays.
- Bump version to 0.6.1

## Additional Notes:

There are situations where many layers of stacked overlay backgrounds may impact the visibility of the site negatively.  These situations should be few and can be revisited from a design language perspective in the future if warranted.

### Docs:

Screen Shot 2022-12-14 at 3 41 35 PM

### Crude Display Tests:

Screen Shot 2022-12-13 at 4 27 48 PM

Screen Shot 2022-12-13 at 4 27 31 PM

### DIFF
--- a/dev/content/Overlays.vue
+++ b/dev/content/Overlays.vue
@@ -418,6 +418,92 @@ useAppFlasher.warning("Hooray!")</code></pre>
 
         <PropsTable :props="tooltipProps"></PropsTable>
       </ComponentLayout>
+
+      <ComponentLayout
+        class="mt-8"
+        :show-badge="false"
+        title="Stacking Context"
+      >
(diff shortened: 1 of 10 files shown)


profsmallpine (approved):
🤌

PR #25, opened 2023-01-04
refactor/base api interface improvements

steven-klein:
# What this does

- Establishes an interface for an `HttpClient` along with convenience interfaces for common Trails response bodies.
- Refactors `BaseAPI` to fulfill the `HttpClient` interface and avoids exposing the library backing the implementation.
- Refactors `useBaseAPI` to mirror `BaseAPI` methods and implementation.
- Provides helper functions and type guards to aid in error checking Promise rejections.
- Removes the `dataIntercept` option from `RequestOptions` interface.
- Adds an `HttpError` class to replace the `AxiosError` interface that was previously received in promise rejections.
- Adds a `patch` request method to the `HttpClient` interface.
- Loosens the type alias for `QueryParams` and `RequestPayload`.
- Sets the `RequestOptions` argument to optional in all `HttpClient` methods.

# What this does not do

I did not choose to create an HttpClient response interface that includes the response status code.  There did not appear to be enough interest in the group to support this, and its use cases were limited.  The side effect would have been a complex response interface that has properties that are rarely if ever accessed after the promise is resolved.  Making the majority of cases more cumbersome than necesary.

# Why

In short, we've been annotating return values and promise results incorrectly when we used BaseAPI.  Let's see how.

__api/users.ts__  

This is a fairly common pattern where we've wrapped up the `BaseAPI.get` method in `UsersAPI.get` and we've annotated that the return value is an `AxiosPromise`.  This is our most common mistake.

```ts
import { AxiosPromise } from "axios"
import { BaseAPI } from "@xy-planning-network/trees"
import { RequestOptions } from "@xy-planning-network/trees/types/api/base"

const UsersAPI = {
  get(opts?: RequestOptions): AxiosPromise {
    return BaseAPI.get("/users", opts || {})
  },
}
```

If we remove `AxiosPromise` and let inference do the work for us, we can see that we're actually getting `Promise<any>`.

Screenshot 2023-01-05 at 5 22 11 PM

So why doesn't Typescript flag this?  The `AxiosPromise` type definition simply extends `Promise`, so they overlap!  But there is one important distinction.  `AxiosPromise` wraps the `Promise<T>` generic in an `AxiosResponse<T>`.  Where `<T>` is the expected shape of the response body.

```ts
export interface AxiosPromise<T = any> extends Promise<AxiosResponse<T>> {}

// notice that T is mapped to the data property of the AxiosResponse interface
export interface AxiosResponse<T = any, D = any>  {
  data: T;
  status: number;
  statusText: string;
  headers: AxiosResponseHeaders;
  config: AxiosRequestConfig<D>;
  request?: any;
}
```

This exposes two problems.  The first is that without declaring the shape of T the result variable of our resolved promise becomes `AxiosResponse<any, any>`.  This means we've effectively disabled type checking on the data property, so typescript will let us attempt to access anything we want on `result.data`.  

The second issue is that, `BaseAPI` doesn't actually return the full `AxiosResponse`.  It returns only the value of the data property and throws the rest away.

```ts
UsersAPI.get()
  .then((result) => {
    // Typescript is happy to let you do this even though headers and status are undefined properties
    const { data, headers, status } = result

    // Typescript will also let you do this because data: any
    // I'm not saying we shouldn't have an interface with a fuManchu property, but definitely don't yet.
    const { id, fuManchu } = data
  })
  .catch((e) => console.log(e))
```

### Why has this been working?

Because all of our responses from Trails are in a shape that includes a top level data property.  Which is a property available on `AxiosResponse`, and luckily we've not been caught out attempting to access a property on data that doesn't exist on the expected result.  I chalk this up to how client side dev work is done, with real time feedback and adjusting according to those runtime errors vs. depending on Typescript to point out issues for us.

```ts
// This is how our responses come in.
interface TrailsResponse<T> {
    data: T
}
```

### What should we be doing

We should do two things.  The first is to stop using `AxiosPromise` to annotate return values and instead allow inference to do the work for us, which is a good practice in general.  Let inference work, it keeps the code cleaner and easier to read and ensures you have the actual return value, provided that value can be inferred.

The second is that we should leverage type generics to define the expected shape of our response payload and avoid results that are typed as `any`.  This will save us grief down the road.  It's much easier to update an interface, run the type check to find the failed property access and update them accordingly, than it is to try and search and replace or update.  It should also speed up dev workflows, inside your `then` block your result variable will be ready to auto complete any properties declared on your interface and ensure that when you are assigning it to an existing variable outside the scope of that block, that it is a valid assignment.  Think `const user = ref<User>()` and `user.value = result.data`.

### How do we update for this change?

It's actually pretty straightforward.  Every `BaseAPI` accepts a generic type definition, you just need to use it.  The hardest part is just ensuring you declare the correct response interface there.  Most commonly, you'll need to wrap it in a `TrailsResponse<T>` interface when your request is hitting our own API endpoints.

```ts
import {
  BaseAPI,
  RequestOptions,
  TrailsResponse,
} from "@xy-planning-network/trees"
import User from "@/types/user"

const UsersAPI = {
  get(opts?: RequestOptions) {
    return BaseAPI.get<TrailsResponse<User>>("/users", opts)
  },
}
```

Typescript is now ready to be helpful.  Sorry, no fuManchu for you.

Screenshot 2023-01-05 at 6 07 08 PM

### Okay, what about error handling

As discussed, we'll take the pragmatic approach to annotating promise rejections as `HttpError | Error` - this process would be known as "type narrowing" which is still much better than accepting `any`.  If you are interested enough, you can deep dive into Typescripts decision to mark `try/catch` error variables as `unknown` but leave promise rejections as `any`.

Provided the response code is 2xx and the response payload matches the definition we provided, this shouldn't error.

```ts
import { ref } from "vue"
import {
  BaseAPI,
  HttpError,
  RequestOptions,
  TrailsResponse,
} from "@xy-planning-network/trees"
import User from "@/types/user"

const UsersAPI = {
  get(opts?: RequestOptions) {
    return BaseAPI.get<TrailsResponse<User>>("/users", opts)
  },
}

const id = ref<number>(0)

UsersAPI.get()
  .then((result) => {
    id.value = result.data.id
  })
  .catch((e: HttpError | Error) => {
    console.log(e)
  })
```

Let's imagine we received a 401 unauthorized status code, and we want to check against that:

```ts
UsersAPI.get()
  .then((result) => {
    id.value = result.data.id
  })
  .catch((e: HttpError | Error) => {
    if (hasHttpErrStatus(e, 401)) {
      // do something meaningful here
    }
    
    console.log(e)
  })
```

If you want to access the status and response properties on the `HttpError` use the available type guard `isHttpError`.  This is a user defined type guard.  It appears to return a boolean, but what it actually does is assert that `e` is in fact an `HttpError` and inside the scope of the conditional block you can safely access the properties available on `HttpError`.  Outside that block, `e` may still be an `Error` type, so `status` access is prohibited.

```ts
UsersAPI.get()
  .then((result) => {
    id.value = result.data.id
  })
  .catch((e: HttpError | Error) => {
    if (isHttpError(e)) {
      // This is an HttpError!
      console.log(e.status)
    }

    console.log(e)
  })
```

Most errors will be `HttpErrors`, but we can't safely assume that all will be.  This is especially true in more complex promise chains where somewhere up the chain a rejection that is not, when we don't leverage type generics and access a property on a result that doesn't exist, or when a response interface changes on the server, but the client interface isn't aware.

```ts
UsersAPI.get()
  .then((result) => {
    const { fuManchu } = result.data
  })
  .catch((e: HttpError | Error) => {
    // Error: fuManchu does not exist
    console.log(e)
  })
```

## How to review this:

- Start by looking through the interface and type declarations in `src/api/client.ts`, specifically the `HttpClient` interface.  This should help you get a baseline for what working with BaseAPI will feel like.  You should notice very little has changed there.
- Next look closely at the additional convenience interfaces like `HttpPromise`, `TrailsPromise`, `TrailsResponse` etc. It may be hard to make sense of them at first, since using generic has not been a common practice, but most we're pretty heavily thought out or derived while updating existing apps.  Now is a good time to nitpick names and descriptions for clarity and intuitiveness.
- Jump into the implementation of BaseAPI, again not much has actually changed here, but inspect how it handles errors from axios and note the addition of 3 functions, two helpers and a type guard.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,6 @@ module.exports = {
     sourceType: "module",
   },
   rules: {
-    "@typescript-eslint/no-non-null-assertion": "warn",
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/no-empty-interface": [
       "error",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,13 @@ module.exports = {
     sourceType: "module",
   },
   rules: {
-    "@typescript-eslint/no-non-null-assertion": 0,
+    "@typescript-eslint/no-non-null-assertion": "warn",
     "@typescript-eslint/no-explicit-any": 0,
+    "@typescript-eslint/no-empty-interface": [
+      "error",
+      {
+        allowSingleExtends: true,
+      },
+    ],
   },
 }
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -1,6 +1,7 @@
 import axios, { AxiosResponse, AxiosRequestConfig } from "axios"
 import {
   HTTP_CANCELLED_ERROR,
+  HTTP_ERROR,
   HttpClient,
   HttpError,
   HttpPromise,
@@ -57,7 +58,7 @@ export const httpRequest = <T = any>(
           err.message,
           err.response?.status,
           err.response?.data,
-          axios.isCancel(err) ? "HttpCanceledError" : undefined
+          axios.isCancel(err) ? HTTP_CANCELLED_ERROR : HTTP_ERROR
         )
       }
 
@@ -86,23 +87,6 @@ const BaseAPI: HttpClient = {
   },
   delete<T = any>(path: string, opts?: RequestOptions) {
     return httpRequest<T>({ url: path, method: "DELETE" }, opts || {})
-  },
-  hasErrStatus(err: unknown, codes: number | number[]) {
-    if (typeof codes === "number") {
-      codes = [codes]
-    }
-
-    if (this.isHttpError(err)) {
-      return codes.includes(err.status)
-    }
-
-    return false
-  },
-  isHttpCancel(err: unknown): boolean {
-    return this.isHttpError(err) && err.name === HTTP_CANCELLED_ERROR
-  },
-  isHttpError<T>(err: unknown): err is HttpError<T> {
-    return err instanceof HttpError
   },
   patch<T = any>(
     path: string,
@@ -135,3 +119,40 @@ const BaseAPI: HttpClient = {
 }
 
 export default BaseAPI
+
+/**
+ * A convenience method for checking if a variable in a failed request has an http status code.
+ * This is most useful for checking for specific http status codes in error callbacks.
+ * @param err unknown
+ * @param codes number | number[]
+ * @returns boolean
+ */
+export const hasHttpErrStatus = (err: unknown, codes: number | number[]) => {
+  if (typeof codes === "number") {
+    codes = [codes]
+  }
+
+  if (isHttpError(err)) {
+    return codes.includes(err.status)
+  }
+
+  return false
+}
+
+/**
+ * A type guard for checking if a variable is in the shape of a HttpError
+ * @param err unknown
+ * @returns err is HttpError
+ */
+export const isHttpError = <T>(err: unknown): err is HttpError<T> => {
+  return err instanceof HttpError
+}
+
+/**
+ * A convenience method for checking if a variable is a cancelled http request error.
+ * @param err unknown
+ * @returns boolean
+ */
+export const isHttpCancel = (err: unknown): boolean => {
+  return isHttpError(err) && err.name === HTTP_CANCELLED_ERROR
+}
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -121,25 +121,6 @@ const BaseAPI: HttpClient = {
 export default BaseAPI
 
 /**
- * A convenience method for checking if a variable in a failed request has an http status code.
- * This is most useful for checking for specific http status codes in error callbacks.
- * @param err unknown
- * @param codes number | number[]
- * @returns boolean
- */
-export const hasHttpErrStatus = (err: unknown, codes: number | number[]) => {
-  if (typeof codes === "number") {
-    codes = [codes]
-  }
-
-  if (isHttpError(err)) {
-    return codes.includes(err.status)
-  }
-
-  return false
-}
-
-/**
  * A type guard for checking if a variable is in the shape of a HttpError
  * @param err unknown
  * @returns err is HttpError
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -104,6 +104,13 @@ const BaseAPI: HttpClient = {
   isHttpError(err: unknown): err is HttpError {
     return err instanceof HttpError
   },
+  patch<T = any>(
+    path: string,
+    data: Record<string, unknown> | FormData,
+    opts?: RequestOptions
+  ) {
+    return httpRequest<T>({ url: path, data, method: "PATCH" }, opts || {})
+  },
   post<T = any>(
     path: string,
     data: Record<string, unknown> | FormData,
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -78,15 +78,15 @@ export const httpRequest = <T = any>(
 }
 
 const BaseAPI: HttpClient = {
+  delete<T = any>(path: string, opts?: RequestOptions) {
+    return httpRequest<T>({ url: path, method: "DELETE" }, opts || {})
+  },
   get<T = any>(
     path: string,
     opts?: RequestOptions,
     params?: Record<string, unknown>
   ) {
     return httpRequest<T>({ url: path, method: "GET", params }, opts || {})
-  },
-  delete<T = any>(path: string, opts?: RequestOptions) {
-    return httpRequest<T>({ url: path, method: "DELETE" }, opts || {})
   },
   patch<T = any>(
     path: string,
--- a/src/api/base.ts
+++ b/src/api/base.ts
@@ -1,20 +1,20 @@
 import axios, { AxiosResponse, AxiosRequestConfig } from "axios"
 import {
-  HTTP_CANCELLED_ERROR,
-  HTTP_ERROR,
+  HTTP_ERR_CANCEL,
+  HTTP_ERR,
   HttpClient,
   HttpError,
   HttpPromise,
-  RequestOptions,
+  ReqOptions,
 } from "./client"
 import { useAppSpinner } from "@/composables/useSpinner"
 
 /**
- * apiDelayIntercept delays the request of an api call by the configured withDelay value in RequestOptions
- * @param config RequestOptions
- * @returns RequestOptions
+ * apiDelayIntercept delays the request of an api call by the configured withDelay value in ReqOptions
+ * @param config ReqOptions
+ * @returns ReqOptions
  */
-const apiDelayReqIntercept = (config: RequestOptions & AxiosRequestConfig) => {
+const apiDelayReqIntercept = (config: ReqOptions & AxiosRequestConfig) => {
   const delay = config.withDelay || 0
   return new Promise((resolve) => {
     setTimeout(() => {
@@ -34,9 +34,9 @@ apiAxiosInstance.interceptors.request.use(apiDelayReqIntercept)
 
 export const httpRequest = <T = any>(
   config: AxiosRequestConfig,
-  opts: RequestOptions
+  opts: ReqOptions
 ): HttpPromise<T> => {
-  const options: RequestOptions = {
+  const options: ReqOptions = {
     skipLoader: false,
     withDelay: 0,
     ...opts,
@@ -58,7 +58,7 @@ export const httpRequest = <T = any>(
           err.message,
           err.response?.status,
           err.response?.data,
-          axios.isCancel(err) ? HTTP_CANCELLED_ERROR : HTTP_ERROR
+          axios.isCancel(err) ? HTTP_ERR_CANCEL : HTTP_ERR
         )
       }
 
@@ -78,12 +78,12 @@ export const httpRequest = <T = any>(
 }
 
 const BaseAPI: HttpClient = {
-  delete<T = any>(path: string, opts?: RequestOptions) {
+  delete<T = any>(path: string, opts?: ReqOptions) {
     return httpRequest<T>({ url: path, method: "DELETE" }, opts || {})
   },
   get<T = any>(
     path: string,
-    opts?: RequestOptions,
+    opts?: ReqOptions,
     params?: Record<string, unknown>
   ) {
     return httpRequest<T>({ url: path, method: "GET", params }, opts || {})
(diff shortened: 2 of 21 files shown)


.eslintrc.js
```diff
@@ -19,7 +19,12 @@ module.exports = {
     sourceType: "module",
   },
   rules: {
-    "@typescript-eslint/no-non-null-assertion": 0,
```
steven-klein, line 22: I had a feeling we weren't triggering this rule and we weren't time to remove it so we don't start skipping it unintentionally.

.eslintrc.js
```diff
@@ -19,7 +19,12 @@ module.exports = {
...
     "@typescript-eslint/no-explicit-any": 0,
+    "@typescript-eslint/no-empty-interface": [
+      "error",
+      {
+        allowSingleExtends: true,
```
steven-klein, line 26: There's value in my opinion in creating what is effectively an interface alias in a library.  Especially when generic are involved and allowing for meaningful names on interfaces that may later be extended.

src/api/base.ts
```diff
@@ -68,58 +32,127 @@ const apiAxiosInstance = axios.create({
...
+
+      window.clearTimeout(wait)
+    })
+}
+
+const BaseAPI: HttpClient = {
   get<T = any>(
     path: string,
-    opts: RequestOptions,
+    opts?: RequestOptions,
```
steven-klein, line 83: I felt like this should be optional or have a default `{}`.  Consumers shouldn't need to provide an empty object as a default.

src/api/base.ts
```diff
@@ -68,58 +32,127 @@ const apiAxiosInstance = axios.create({
...
 export default BaseAPI
+
+/**
+ * A convenience method for checking if a variable in a failed request has an http status code.
+ * This is most useful for checking for specific http status codes in error callbacks.
+ * @param err unknown
+ * @param codes number | number[]
+ * @returns boolean
+ */
+export const hasHttpErrStatus = (err: unknown, codes: number | number[]) => {
```
steven-klein, line 130: Some amount of type narrowing should be used in catch blocks in favor of accepting `any`.  This makes it nice an easy to check an unknown variable for a status property.  `err?.status === 409` is okay, but will only pass type checks with an implicit/explicit any or an overly narrow assertion of `HttpError`.  Let's assert `HttpError | Error` and use this method for convenience when checking for status codes.

src/api/base.ts
```diff
@@ -68,58 +32,127 @@ const apiAxiosInstance = axios.create({
...
+
+  return false
+}
+
+/**
+ * A type guard for checking if a variable is in the shape of a HttpError
+ * @param err unknown
+ * @returns err is HttpError
+ */
+export const isHttpError = <T>(err: unknown): err is HttpError<T> => {
```
steven-klein, line 147: Note the return annotation is `err is HttpError`.  This is your user defined type guard.  It's like using `typeof err === "string"`, but for user defined interfaces.  If the condition is truthy, then the variable inside that block will be inferred to be of type `HttpError`.

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
+/**
+ * RequestMethod
+ * The HTTP request methods that our http client supports.
+ */
+export type RequestMethod =
+  | "GET"
+  | "get"
+  | "PATCH"
```
steven-klein, line 8: Patch was just an easy addition.  It may not have a use case yet.

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
...
+ * HttpPromise
+ * The successfully resolved interface of an http client request.
+ */
+export interface HttpPromise<T = any> extends Promise<T> {}
+
+/**
+ * TrailsPromise
+ * The successfully resolved interface of an http client request returned by @xy-planning-network/Trails.
+ */
+export interface TrailsPromise<T = any> extends Promise<TrailsResponse<T>> {}
```
steven-klein, line 68: The first of many convenience interfaces.  It saves you from the problem of deeply nested generics (feedback from portal implementations):

```ts
HttpPromise<TrailsResponse<User>>> === TrailsPromise<User>
```

Currently, the Tax app requires functions to annotate their return value, which is a typescript best practice, but is only covered by eslint and is not a Typescript rule.  Jury is out on whether that eslint rule should be used elsewhere.  Thee Tax app uses the `TrailsPromise` interface to annotate it's api calls.

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
...
+ *     .then(result => {
+ *         const currentPage = result.data.page
+ *         const users = { ...result.data.items }
+ *
+ *         result.data.items.forEach(u => {
+ *             console.log(u.id)
+ *         })
+ *     })
+ */
+export interface TrailsResponsePaged<T = any> {
```
steven-klein, line 118: There are a lot of different ways to declare these "paged" interfaces, but this was the easiest on the eyes for me.

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
...
+ * QueryParams
+ * Http GET request query params.
+ */
+export type QueryParams = Record<string, any>
+
+/**
+ * RequestPayload
+ * Http POST and PUT payloads.
+ */
+export type RequestPayload = Record<string, any> | FormData
```
steven-klein, line 138: `RequestPayload` and `QueryParams` have been loosened to `Record<string, any>`.  It's been cumbersome to read things like `UserAPI.update({...user})` or `UserAPI.update(user as Record<string, unknown>)`.

Now nice and easy:
`UserAPI.update(user)`

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * HttpClient
+ * The http client interface the BaseAPI implements.
+ */
+export interface HttpClient {
```
steven-klein, line 144: There's nothing really new here.  BaseAPI pretty much already fulfilled this interface, but it was time to define it explicitly.  This is not a particularly onerous interface to fulfill, we should be able to swap out implementations, and we may want to define arguments or even Vue component props that accept any implementation of our HttpClient interface.

src/entry.ts
```diff
@@ -24,5 +38,16 @@ export * from "@/composables/index"
 // each can be registered via Vue.component()
 export * from "@/lib-components/index"
 
-export { BaseAPI }
-export type { RequestOptions }
+// Http Client exports
+export { BaseAPI, hasHttpErrStatus, isHttpCancel, isHttpError }
+export type {
```
steven-klein, line 43: You'll start to see this a lot more.  Seems to be a hallmark of better libraries to export types that are expected to be consumed.  Vue for example lets you `import {Ref, ComputedRef, etc...} from "vue"`.  Trees should do the same, so you don't need to import from a nested `types` directory.

src/lib-components/lists/Table.vue
```diff
@@ -96,8 +96,12 @@ const loadAndRender = (): void => {
     q: query.value,
   }
 
-  BaseAPI.get(props.tableData.url, { skipLoader: !props.loader }, params).then(
-    (success: AxiosResponse) => {
+  BaseAPI.get<TrailsResponsePaged<unknown>>(
```
steven-klein, line 99: The reason `Trails` prefixed interfaces ultimately made it into Trees.  Because Trees loves Trails and Trails love Trees.

DavidLarsKetch (approved):
First off, thank you for the extensive writeup! Laying out the problems at hand with typing was very helpful. It is hard for me to appreciate the nuances, but they've impacted the dev workflow and more strictly typed setup may mitigate all of those. That being said, I can't say much about improvements to the change to the API since the impact is more on the response handling side, so there may be feedback then if/when the ergonomics miss the mark.

Secondly, this enables fruitful debate about the shape of the interaction between trees and trails now that both can largely be updated in one place. Administratively, my gut is that debate should be had now before fully implementing this PRs changes in our apps, especially since the shape of `TrailsResponse` is open for debate. Thataways, those are one-time changes. Maybe something to dig into at the retreat.

src/api/base.ts
profsmallpine: I'm thinking about a dev 2 years from know looking at this function and curious why we aren't doing `err?.status === 409`. I'm not sure how to expose this info, but i'd love to see this live with the code in some way

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
+/**
+ * RequestMethod
+ * The HTTP request methods that our http client supports.
+ */
+export type RequestMethod =
```
profsmallpine, line 5: In implementing, I found some of the line lengths on these things to get in the way. I'm curious how we @DavidLarsKetch @steven-klein feel about change `Request` to `Req` and `Response` to `Resp` for brevity sake.

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
...
+ *     .then((result) => {
+ *         const currentPage = result.data.page
+ *         const users = result.data.items
+ *
+ *         users.forEach((u) => {
+ *             console.log(u.id)
+ *         })
+ *     })
+ */
+export interface TrailsResponsePaged<T = any> {
```
profsmallpine, line 118: e.g.`TrailsRespPaged`.

I appreciate the naming parity with trails across this stuff. `Paged` matching the struct in trails that is.

profsmallpine (approved):
Getting down to the most minor of notes at this point, looking good!

src/api/base.ts
steven-klein: Yeah, this is a generally unpopular design decision in TypeScript it seems if you look at the number of 👎 tags on this [comment](https://github.com/microsoft/TypeScript/issues/33736).

One alternative, which may be better for clarity, is to use the type guard.  

```ts
UsersAPI.get(id)
  .then((result) => {
    user.value = result
  })
  .catch((err: HttpError | Error) => {
    if (isHttpError(err) && err.status === 409) {
      // do conflict handling
    }
  })
```

I think we had a few hairy ones on the portal side that pushed for more convenience and readability.
```ts
    if (isHttpError(err) && (err.status === http.statuses.BAD_REQUEST || http.statuses.INTERNAL_SERVER_ERROR)) {
      // do conflict and badhandling
    }
```
But there are many other ways to clean this up beyond a helper method like `hasHttpErrStatus`.

steven-klein: I should note the `isHttpError` type guard was directly inspired by [Axios](https://github.com/axios/axios#typescript).

src/api/client.ts
```diff
@@ -0,0 +1,197 @@
...
+ */
+export interface HttpClient {
+  /**
+   * The method to make an http GET request.
+   * @param path string
+   * @param opts RequestOptions
+   * @param params QueryParams
+   * @returns HttpPromise<T>
+   */
+  get<T>(
```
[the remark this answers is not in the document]
steven-klein: That's my usual pref as well.  Good catch.

src/api/client.ts
steven-klein: @DavidLarsKetch seems like a pretty solid idea to me.  Give a thumbs up if you're onboard.  I'm happy to implement in a separate PR, so we can see it.

src/api/client.ts
steven-klein: `HttpPromise` is effectively a copycat of `AxiosPromise`, but since trees doesn't have any kind of response schema and just forwards the response body, it ends up being a straight overlap of `Promise`.  I could see a day when it's time to wrap all responses in some kind of schema that would include the response code, headers, and data etc., which is why I've defined it early.  When/if it does happen, Tree's consumer can update, type check and know exactly where they need to update their access to the response body.

At the moment, the only use for `HttpPromise` would be if we need to hit a 3rd party API endpoint.

PR #26, opened 2023-01-06
RadioCards: when disabled, has a regular cursor

DavidLarsKetch:
Opted for the regular cursor instead of `cursor-not-allowed`. Personally, not a huge fan of that cursor style and encourages someone to think something's broken or not in an expected state.

### DIFF
--- a/src/lib-components/forms/RadioCards.vue
+++ b/src/lib-components/forms/RadioCards.vue
@@ -118,7 +118,7 @@ const nameAttr = computed(() => {
           :class="[
             checked ? 'border-transparent' : 'border-gray-300',
             active ? 'border-blue-500 ring-2 ring-blue-500' : '',
-            disabled ? 'cursor-auto opacity-75' : '',
+            disabled ? 'cursor-not-allowed opacity-75' : '',
           ]"
         >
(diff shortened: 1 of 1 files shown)


steven-klein (changes requested):
I have to admit I've never really liked `cursor-not-allowed` as it feels unintuitive to me, but at the moment it looks like all or at least most of our form elements are using it as their disabled state.  For the sake of consistency, we should use it here as well.

FWIW, this appears to be the suggested disabled cursor on Tailwind UI's form components.  There's a good chance that's where the pattern crept in from and because I lack an informed opinion on how beneficial it is or isn't to users, I'm going to punt to - they're doing it, so we should too.  For now anyway.

steven-klein:
Since the changes here are relatively small, I've set this to merge into #release/0.6.2.  Feels like the current use case that kicked this off can move forward without it and be further improved by this on release.

steven-klein:
@DavidLarsKetch I'm going to push this forward with `cursor-not-allowed` which I know neither of us love, but it's consistent with our other form elements.  Happy to revisit alternative cursor's later in the context of all form elements.

PR #30, opened 2023-04-14
Maint/dependencies

steven-klein:
# What this does

- bumps .node-version to 18.15.0
- updates github action dependencies and adds action for build test on feature branches
- updates all build and dev dependencies to latest stable versions
- adds shared eslint and prettier configs in the configs directory
- bumps vue lint rules from essential to recommended
- updates application wrapper props to handle props formatting expectations for vue
- includes minor bug fixes

## Notes

As discussed, bumping prettier means the class attribute will be [collapsed into a single line](https://prettier.io/blog/2021/11/25/2.5.0.html#collapse-html-class-attributes-onto-one-line-11827httpsgithubcomprettierprettierpull11827-by-jlongsterhttpsgithubcomjlongster) going forward.  The Tailwind has made their [personal preferences known](https://github.com/prettier/prettier/issues/10918#issuecomment-851049185), so it's unlikely to change anytime soon, and we should more forward.  We can revisit our options once everyone has had a chance to feel it out.

After working with [Tailwinds auto class sorting prettier plugin](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier) briefly, I'm punting on integrating it for the time being.  The DX was painfully [slow](https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/144) triggering on save and since it's not something we've been using, I don't feel a rush to integrate it.  Prettier plugins are autoloaded and wouldn't require specific configs changes to Trees, so any existing project can begin integrating with `prettier-plugin-tailwind` at anytime.


.github/workflows/publish.yml
```diff
@@ -11,26 +11,24 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
         with:
-          path: client/node_modules
-          key: npm-cache-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            npm-cache-
+          node-version-file: ".node-version"
+          cache: "npm"
+          cache-dependency-path: ./package-lock.json
```
profsmallpine, line 22: 🥳

package.json
```diff
@@ -34,32 +34,28 @@
   },
   "devDependencies": {
     "@point-hub/vue-highlight": "^0.1.4",
+    "@rushstack/eslint-patch": "^1.2.0",
```
profsmallpine, line 37: 🤩

src/lib-components/layout/SidebarLayout.vue
```diff
@@ -18,20 +18,20 @@ import { ref } from "vue"
 
 const props = withDefaults(
   defineProps<{
-    activeURL?: string
-    iconURL: string
+    activeUrl?: string
```
profsmallpine, line 21: Nitpick: go is telling us to do activeURL or userID. Specify a rule to keep the ts/go friendship? Not a huge deal..

profsmallpine (approved):
noice cleanup here

DavidLarsKetch (approved):

dev/content/Home.vue
```diff
@@ -184,7 +184,7 @@ const features = [
                   >
                     <img
                       class="h-12 w-12 rounded-full"
-                      src="https://4e5jbr2e8ha51iaeaa2xu4qj-wpengine.netdna-ssl.com/wp-content/uploads/2020/11/Jillian-Berry.png"
+                      src="https://content.malakye.com/media/uploads/images/personalprofiles/thumbs/742194_637214344006622155.png"
```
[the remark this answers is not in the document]
steven-klein: It's actually still an image of Jillian!  The old URL was just returning 404.

src/lib-components/layout/SidebarLayout.vue
steven-klein: As far as these specific app layout components go, I'd probably just suggest we plan to change the prop names.  The icon prop can probably move to `iconSrc` and I'm not sure if we should have an `activeURL` prop at all.  Will it ever be anything other than what can be read from window.location?

profsmallpine: No prob, not worth an ongoing battle

profsmallpine: Yeah better to use window.location for this

PR #31, opened 2023-04-14
chore: lint and format

steven-klein:
A companion PR to #30 focused on linting and formatting fixes

# What this does
- removes custom rules from eslint and resolves any remaingin issues
- adds a `format` script for cli usage
- auto fixes with eslint
- auto formats with prettier


profsmallpine (approved):
🚀

src/lib-components/overlays/Flash.vue
```diff
@@ -1,4 +1,7 @@
 <script setup lang="ts">
+// TODO: track down instances and see if this is really needed
+/* eslint-disable vue/no-v-html */
```
DavidLarsKetch, line 3: As far as I know, `college-try` is the only app leveraging this. There are [2 flashes available](https://github.com/xy-planning-network/college-try/blob/07fb64950845808e731af5aa190a1f9c19df60f5/http/web/web.go#L63-L98) that have HTML in them. They are only used in the [web/login_handler](https://github.com/xy-planning-network/college-try/blob/master/http/web/login_handler.go) for a grand total of 4 instances.

XSS is a possibility insofar as the `email` param for those functions comes from user input in a form. Now, how it passes other guards - including those in [html/template](https://pkg.go.dev/html/template) - sounds like a sophisticated attacker.

DavidLarsKetch (approved):
If my comment helps resolve the TODO, then definitely knock that out before merging, but LGTM otherwise.

src/lib-components/overlays/Flash.vue
steven-klein: Yeah, I think the real surface area of that component as an attack vector remains reasonably small.  If there was a one or two line refactor, I'd attempt to resolve it, but the fix likely either needs to be extending the API of the component and associated composable and/or updating the app's usage, so we'll backlog it for now.

PR #33, opened 2023-04-24
Spk/tables refactor

steven-klein:


profsmallpine:
Couple minor thoughts:

1. `event` in the `TableActionItem` felt out of touch after this change, naming it `click` or `onClick` is more intuitive for me
2. having the `tableActionsType` as it's own prop felt a little cumbersome. was thinking it should live in the `tableOptions` but realized that `DataTable` doesn't have options. no solution jumps to mind, but if this bothered you at all, maybe we can re-org it

steven-klein:
> Couple minor thoughts:
> 
> 1. `event` in the `TableActionItem` felt out of touch after this change, naming it `click` or `onClick` is more intuitive for me
> 2. having the `tableActionsType` as it's own prop felt a little cumbersome. was thinking it should live in the `tableOptions` but realized that `DataTable` doesn't have options. no solution jumps to mind, but if this bothered you at all, maybe we can re-org it

1. (+1) for renaming it to `onClick`.
2. It does bother me a bit.   One question might be whether there's value in making the `type` of actions you want a required definition.  Which is a tradeoff between the convenience of a default for clarity and intention.  

In that case, the props could look like this:
```ts
type TableActions<T> = {
  type: "dropdown" | "buttons",
  actions: TableActionItem<T>[]
}

const props = withDefaults(
  defineProps<{
    tableActions?: TableActions<any>
    tableColumns: TableColumns<any>
    tableOptions: DynamicTableOptions
  }>(),
  {
    tableActions: () => ({type: "dropdown", actions: []}),
  }
)
```

src/composables/useTable.ts
```diff
@@ -0,0 +1,107 @@
...
+    console.warn(
+      "refresh() was called on a static table, did you mean to use DynamicTable?"
+    )
+  },
+  reset() {
+    console.warn(
+      "reset() was called on a static table, did you mean to use DynamicTable?"
+    )
+  },
+}
```
DavidLarsKetch, line 24: Clever ⭐

DavidLarsKetch:
>  One question might be whether there's value in making the type of actions you want a required definition.

I vote for making it required. I had a little trouble tracking the props just now when comparing the changes in this PR to the draft PRs in the other repos for the table components. Each of the other props are complicated (i.e., require thinking about how your object translates to a tabular view) and I'd just as likely overlook the `tableActionsType` prop. Not sure if that totally makes sense, but sounds like we're all on the same page in general.

Said another way, when composing the table actions, to me it seems like a good starting point to force thinking about the UI ("do I want buttons or a dropdown?"), which may inspire critical thinking about overstuffing the cell with actions in the first place (looking at you `/users/Main.vue` on Portal).

DavidLarsKetch (commented):
Looks good overall.

The fact that whether an action is enabled or not is alternatively a boolean or a function is a bit dense to grok the value of as well. Actually, I'm not even sure I've got it all right in my head. Maybe a comment or two highlighting the complexity here, why its important, etc. may help with future maintenance.

PR #37, opened 2023-06-15
integrate floating UI's useFloating composable on Popover, Tooltip, and ActionsDropdown

steven-klein:
While reviewing #36 I noticed that there were a few runtime errors on the popover auto placement strategy from the bump to Vue 3.3, so it seemed like as good a time as any to integrate a 3rd party and solve the table actions dropdown display bug as well.

[Floating-UI](https://floating-ui.com/docs/vue) is the successor to PopperJS and has first class Vue support.  It does add a little heft to the bundle, but the trade-off seems worthwhile.

### DIFF
--- a/src/lib-components/overlays/Popover/Popover.vue
+++ b/src/lib-components/overlays/Popover/Popover.vue
@@ -1,229 +1,59 @@
-<script lang="ts">
-export type PopoverPosition =
-  | "top-left"
-  | "top-center"
-  | "top-right"
-  | "bottom-left"
-  | "bottom-center"
-  | "bottom-right"
-  | "left"
-  | "right"
-  | "auto"
-  | "none"
-</script>
 <script lang="ts" setup>
-import { throttle } from "@/helpers/Throttle"
 import {
   Popover as HeadlessPopover,
   PopoverButton as HeadlessPopoverButton,
   PopoverPanel as HeadlessPopoverPanel,
 } from "@headlessui/vue"
-import { computed, onMounted, onUnmounted, ref } from "vue"
+import { computed, ref } from "vue"
+import {
+  useFloating,
+  offset,
+  autoPlacement,
+  autoUpdate,
+  shift,
+} from "@floating-ui/vue"
+import type { Placement } from "@floating-ui/vue"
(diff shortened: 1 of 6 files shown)


src/lib-components/overlays/Tooltip.vue
```diff
@@ -36,9 +37,9 @@ const hoverPopover = (e: MouseEvent, open: boolean): void => {
 
 <template>
   <Popover :position="position" :as="as">
-    <template #button="{ open, close }">
+    <template #button="{ open, close }: { open: boolean, close: () => void }">
```
steven-klein, line 40: This is just a type annotation for the slot props, which are not currently available via inference with HeadlessUI components.

src/lib-components/overlays/Tooltip.vue
```diff
@@ -36,9 +37,9 @@ const hoverPopover = (e: MouseEvent, open: boolean): void => {
...
       <div
-        class="leading-none w-4 h-4"
+        class="leading-none relative w-4 h-4"
```
steven-klein, line 42: The "large target" hack I added to the tooltip was ill placed under some scenarios, this should keep it well placed on top of the icon.

src/lib-components/overlays/Popover/Popover.vue
```diff
@@ -1,229 +1,59 @@
...
 
 const props = withDefaults(
   defineProps<{
     as?: string
-    position?: PopoverPosition
+    position?: Placement
   }>(),
   {
     as: "div",
-    position: "auto",
```
profsmallpine, line 30: I think the auto string felt more explicit. Maybe Im just used to it at this point, but undefined is not clear at first grok that it is then using auto.

profsmallpine (approved):
Thanks for getting into the weeds on this one :) Just one comment I'll leave up to you whether to change or not

src/lib-components/overlays/Popover/Popover.vue
steven-klein: Yeah, that's a great point from an API clarity standpoint. Updated!

PR #38, opened 2023-06-30
Adds limit handling to DateRangePicker and pass through prop on DynamicTable

profsmallpine:
For naming, I went as explicit as possible on the DateRangePicker and on the DynamicTable, I wanted to keep it as close to `dateSearch` as possible so it's obvious they are related. Considered moving `dateSearch` + `dateSearchMaxRange` into a nested object on the type but this is already nested so skipped that. 

On the implementation, there are a handful of ways to approach, but using maxDate kept the component redraw in the existing logic. I am assuming that a date should be prefilled when using the range, but this could be a separate prop in the future.

### DIFF
--- a/src/lib-components/forms/DateRangePicker.vue
+++ b/src/lib-components/forms/DateRangePicker.vue
@@ -11,13 +11,13 @@ const props = withDefaults(
       minDate: number
       maxDate: number
     }
-    allowedRangeInDays?: number
+    maxRange?: number
     startDate?: number
     label?: string
     help?: string
   }>(),
   {
-    allowedRangeInDays: 0,
+    maxRange: 0,
     startDate: 0,
     label: "",
     help: "",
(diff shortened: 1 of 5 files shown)


src/lib-components/forms/DateRangePicker.vue
```diff
@@ -11,11 +11,13 @@ const props = withDefaults(
       minDate: number
       maxDate: number
     }
+    allowedRangeInDays?: number
```
steven-klein, line 14: nit: I think I'd just go with `maxRange`, which feels intuitive enough for me, given the picker doesn't currently allow time selections.

src/lib-components/forms/DateRangePicker.vue
```diff
@@ -51,7 +53,34 @@ onMounted(() => {
...
+        // Clone date so as to not change selectedDates[0] value
+        var daysAhead = new Date(selectedDates[0].getTime())
+        daysAhead.setDate(daysAhead.getDate() + props.allowedRangeInDays)
+        const now = new Date()
+
+        if (daysAhead > now) {
+          daysAhead = now
+        }
+
+        self.set("maxDate", daysAhead)
```
steven-klein, line 78: I think you'll need to set the `minDate` here is well to prevent backward range selections outside the range bounds.

src/lib-components/forms/DateRangePicker.vue
```diff
@@ -51,7 +53,34 @@ onMounted(() => {
         })
       }
     },
-  })
+  }
+
+  if (props.allowedRangeInDays) {
+    // Set the range to a prefilled value given the allowed range
+    const daysAgo = new Date()
+    opts.defaultDate = [
```
steven-klein, line 61: Watch out for this one.  Looks like you'll end up with what appears to be a completed field, but a `modelValue` that isn't updated, since it's only updated through flatpicker's `onClose` event.

steven-klein (changes requested):
Left a couple of notes to take a look at for some edge cases.

If we find the need to extend the `DateRangePicker` props further, I'd probably make the move to have consuming components like `DynamicTable` accept a single `dateSearch` prop that is a union of `boolean | DateRangeProps`, so that it has the flexibility to touch the configuration without the need to continuously add new props onto `DynamicTable`.

Tricky and a little surprising, Flatpickr doesn't have an option for this in range mode!

PR #42, opened 2023-08-28
bumps axios@1.5.0 and sets config value for params indexes

steven-klein:
# What this does

- bumps axios@1.5.0
- implements axios interface changes
- set's the `paramsSerializer.indexes` option to null on the axios instance used by BaseAPI


profsmallpine (approved):
Looks good, 0.27.2 -> 1.5.0 makes it look like we are 5 years out, but only a little over a year 😅

DavidLarsKetch (approved):

PR #43, opened 2023-08-31
feat: implement brand updates

steven-klein:
# What this does

This update implements the minimum changes expected to stay consistent with incoming brand guidelines:

- adds new brand color palettes and typography
- applies updated colors to app shells, buttons, form inputs, and some select components

## Notes

- The use of `font-display` is entirely optional and should be used sparingly.  It works well for large headings in CTA banners and should typically be set alongside `font-extrabold` with a line-height that is the same or very close to the font size.
- The default color weight of `xy-green` changes from a 600 to a 300 and does not meet accessibility requirements as text.  It should primarily be used as a background color with `text-xy-black` or a similar 800+ color weight.  Some exceptions will apply to icons or other presentation role interface items where legibility is less of a concern.
- The spinning indicator has a heavy branding baked into it.  It has been prepped to deploy alongside logo changes at a later date.  It will for a short period maintain old brand colors and marks.
- Tailwindcss@3.3 now ships with the line-clamp utility baked in, and the additional module is no longer required

## To Preview In An Existing App

`npm install @xy-planning-network/trees@dev`
`npm install --save-dev tailwindcss@latest`

## Breaking Changes

- `xy-btn-green` has been removed
- `xy-navy` has been removed (portal is likely to continue to stub this into its config to ease the refactor)
- `xy-dawn` has been removed
- `xy-lime` has been removed

## Major Considerations

- `xy-green` is a very different default color, best to review existing uses


steven-klein:
@profsmallpine There's a @dev version up if you want to see how this feels in Archive.  I left details in the PR note about setting the neutral palette in the app config as well.  I could use a second opinion on whether not we do a full replacement and tweak as needed, or leave it intact and implement neutral via search and replace or as we see fit here and there.

Ergonomically, it's quite a bit nicer to type `bg-gray`, `text-gray` etc.  Though autocompletion does most of the work anyway on `text-neutral`.

src/lib-components/indicators/XYSpinner.vue
steven-klein, line 1: This is the only component that presented any kind of logistical challenge, since it contains a version of the XY logo.  I applied the new brand colors to the icon, and left a potentially suitable replacement (the new logo mark) in there to be bumped at a later date.  I don't feel particularly compelled to cut another release to coincide with the logo reveal - I think we'd just live with the XY until the next Trees bump comes around.

profsmallpine (approved):

PR #44, opened 2023-09-14
adds branding to spinner indicator, includes inline visual on docs page

steven-klein:
Screenshot 2023-09-14 at 11 21 11 AM

Note: There are some edge cases where multiple layers of background overlay can start to make the spinner difficult to see, but they should be minimal.  If we start to see it frequently, we can revisit.

### DIFF
--- a/dev/content/Overlays.vue
+++ b/dev/content/Overlays.vue
@@ -261,10 +261,15 @@ useAppFlasher.info("Sticky!", true)
           <label class="block text-sm font-medium text-gray-700">
             <ClickToCopy :value="spinnerCopy" />
           </label>
-          <div class="mt-1">
-            <button type="button" class="xy-btn" @click="spinner()">
-              Show Me
-            </button>
+          <div class="space-y-5">
+            <div>
+              <XYSpinner />
+            </div>
+            <div class="mt-1">
+              <button type="button" class="xy-btn" @click="spinner()">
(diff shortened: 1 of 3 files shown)


DavidLarsKetch (approved):

profsmallpine (approved):
This looks good, sorry for hijacking this PR with another question.. but I noticed that on some of my tables where the query is slower, e.g. emails, the loader never pops up. Was this removed from the table? I can dig in but figured you might know of the top of your noggin.

steven-klein:
@profsmallpine that was [unintentional](https://github.com/xy-planning-network/trees/commit/20380e94286b74f26b3ef601efadfa065a7add3a#diff-6661e644a775e6f07ec3301fd7636f894fb568bb103f08b98dd322b9dbbe437eR44). Looks like I just promoted the default prop value, but I think all existing use cases would prefer the loader?

3 in Portal:

- [x] notifications
- [x] users
- [x] invoices

profsmallpine:
@steven-klein yeah on by default is sane to me, especially since it was some hot millis before kicking in

steven-klein:
> @steven-klein yeah on ~by default~ for all is sane to me, especially since it has some hot millis before kicking in

Will include the bump in this PR.  The negation was slippery `{ skipLoader: !props.loader },`!

PR #45, opened 2023-09-25
src/composables/index: exports useBaseAPIPatch

DavidLarsKetch:
🙃 

Moving `PUT /users` over to `PATCH /users` on `college-try`, found this.


steven-klein (approved):
🚀

PR #48, opened 2023-10-06
1. general improvements to form input visibility, props and attrs support

steven-klein:
> PR 1 of 3 related to [Form Input Updates](https://www.notion.so/xylabs/8f16954aa3494675837f0cceb6a21c95?v=9d015d84f6fa48fe837c8ad0f6ab5be8&p=ecae8501d1924fa88b6561c05235759b&pm=s)

# What this does

- improves form input styles (lightens borders, placeholders, labels, and help text)
- improves disabled state (sets light gray background)
- improves error state (ensures borders, placeholders, and focus state's have `classname` precedence)
- adds label and help text props to Toggle
- adds wraps all inputs in a parent div for layout consistency
- adds `useInputField` composable
- adds `error` string prop for setting error state
- removes most default placeholders (date range picker and select will continue to set a default)
- removes `legend` prop on components that used it and normalizes it to `label`
- removes 4 column layout from components that used it (not currently in use - unlikely to be useful)
- removes automatic attribute inheritance from inputs (they are already bound to the input as expected, any additional fall-through inheritance leads to unexpected results in the component and markup)
- removes `.xy-input-error` class (this failed to set some styles, and a lack of attribute inheritance makes a class attribute a difficult method for applying an error state)
- removes input[type] global styling (it was incomplete to begin with, let's try to ensure we are using `BaseInput` - we can easily bring this back if we find it's necessary)
- removes non-text type input support from `BaseInput` (v-model type definitions don't support input types like `file` and there was unnecessary complexity supporting both text like and non-text like inputs - only non text like input I'm aware of is the file uploader on Portal).
- replaces the dynamic table search input with an instance of `BaseInput`

## Notes

- There are a lot of changes in the PR, but it's mostly a matter of, set the pattern in `BaseInput` and apply it to all other inputs.  Spin up the dev server and take a look at the forms page to get a sense of the visual changes.
- The error state has more management and display coming in a later PR
- The Toggle button is a button.  This is noted in the docs now, it shares similarities to inputs but should be thought of as a button and not an input.  Our current usage tracks with this.

### DIFF
--- a/dev/content/Forms.vue
+++ b/dev/content/Forms.vue
@@ -394,8 +394,9 @@ const toggleProps = [{ name: "modelValue", required: true, type: "string" }]
 
             <MultiCheckboxes
               v-model="multiCheckboxSelection"
+              label="Lay it out in a grid"
               :columns="2"
-              help="Set the columns prop to 2, 3, or 4"
+              help="Set the columns prop to 2 or 3"
               :options="options"
             >
               <template #legend>In A Grid Too</template>
@@ -440,7 +441,8 @@ const toggleProps = [{ name: "modelValue", required: true, type: "string" }]
 
             <Radio
               v-model="radioSelection"
-              help="Set the columns prop to 2, 3, or 4"
+              label="Lay it out in a grid"
+              help="Set the columns prop to 2 or 3"
               :options="options"
               :columns="2"
               required
--- a/dev/content/Forms.vue
+++ b/dev/content/Forms.vue
@@ -188,12 +188,16 @@ const toggleProps = [{ name: "modelValue", required: true, type: "string" }]
             <ClickToCopy :value="inputCopy" />
           </label>
           <div class="mt-1">
-            <BaseInput
-              help="No wrong answers here."
-              type="text"
-              label="What's your life moto?"
-              placeholder="It's good to be alive"
-            />
+            <form @submit.prevent>
+              <BaseInput
+                help="No wrong answers here."
+                type="text"
+                label="What's your life moto?"
+                placeholder="It's good to be alive"
+                required
+              />
+              <button type="submit" class="xy-btn">Submit</button>
+            </form>
           </div>
         </div>
 
@@ -203,10 +207,10 @@ const toggleProps = [{ name: "modelValue", required: true, type: "string" }]
           </label>
           <div class="mt-1">
             <BaseInput
+              class="xy-input-error"
               type="text"
               label="Broken"
               placeholder="An invalid input"
-              invalid
             />
           </div>
         </div>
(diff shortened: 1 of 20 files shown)


src/index.css
```diff
@@ -32,17 +32,6 @@
   h6 {
     @apply text-xs leading-4 font-semibold;
   }
-
-  /* Forms: here for backward compatibility.  Use <BaseInput> instead. */
```
profsmallpine, line 36: 😮 ⚡

src/index.css
```diff
@@ -131,9 +120,4 @@
   .max-h-screen-1\/2 {
     max-height: 50vh;
   }
-
-  /* Forms */
-  .xy-input-error {
```
profsmallpine, line 136: 👋

src/lib-components/forms/FieldsetLegend.vue
```diff
@@ -1,23 +1,23 @@
...
-    v-bind="{
-      ...$attrs,
-      class: 'text-sm font-semibold leading-snug text-gray-900',
-    }"
+    v-if="label"
+    class="block text-base leading-snug font-medium text-gray-800"
+    v-bind="$attrs"
   >
-    <slot />
+    {{ label }}
```
profsmallpine, line 21: slot 👋

profsmallpine (approved):

dev/content/Forms.vue
```diff
@@ -204,20 +207,31 @@ const toggleProps = [{ name: "modelValue", required: true, type: "string" }]
           </label>
           <div class="mt-1">
             <BaseInput
+              class="xy-input-error"
```
DavidLarsKetch, line 210: I notice the addition here and deletion in `index.css` Is the idea to use the `error` prop instead? Asking more from a usage standpoint and groking the changes rather than a nitpick.

src/lib-components/forms/BaseInput.vue
```diff
@@ -1,87 +1,73 @@
...
+  | "week"
+
+withDefaults(
   defineProps<{
-    type: string
+    type: TextLikeInputs
     help?: string
     label?: string
     modelValue?: string | number
+    error?: string
```
DavidLarsKetch, line 30: So, just to be clear, the reason this is `string` now is how in #50 the POC displays the actual message passed along? Same question for the other inputs.

DavidLarsKetch (commented):
Still working my way through these, will have more thoughts tomorrow!

dev/content/Forms.vue
steven-klein: Right, the intention is to set the error prop instead of using the class.  That was just left behind in the docs, but is cleaned up now.  The class based styling didn't correctly apply the styles it defined to begin with and had limited utility in general as it was really only suitable for text like inputs.

src/lib-components/forms/BaseInput.vue
steven-klein: Yep, once you go down the path of choosing to set error styles based on the invalid event, you need to bring your own error message display.  The browser typically only displays its tooltip error message on the first input to contain an error, and removes it once the input is in the focus state.  Additionally, it only shows that first error, so it may take a user multiple submission attempts to learn they have more than 1 error in the form.

PR #49, opened 2023-10-10
2. extract input types, normalize props as much as possible

steven-klein:
> PR 2 of 3 related to [Form Input Updates](https://www.notion.so/xylabs/8f16954aa3494675837f0cceb6a21c95?v=9d015d84f6fa48fe837c8ad0f6ab5be8&p=ecae8501d1924fa88b6561c05235759b&pm=s)

# What this does

- Attempts to normalize form input props as much as possible and extracts types as needed (this is part of the broader goal of developing a form schema and simplifying new input development)
- Adds generic to `RadioCards` option type to help support type inference inside the `sublabel` slot (primarily support Portal's use case with benefits)
- Establishes that v-model is an optional prop across all inputs (this seems to be desirable from a flexibility standpoint, and as we dig further into partial updates)
- Bumps vue-tsc@latest to get generic component support

## Notes

- PR 3 catches a small handful of inconsistencies in the way events and attributes are bound to inputs

### DIFF
--- a/dev/content/Forms.vue
+++ b/dev/content/Forms.vue
@@ -1,77 +1,18 @@
 <script setup lang="ts">
 import { ref } from "vue"
+import type { InputOption } from "@/composables/forms"
 
-const commonProps = [
-  { name: "label", required: false, type: "string" },
-  { name: "help", required: false, type: "string" },
-]
-const isChecked = ref(true)
-const checkboxCopy = `<Checkbox label="I'm here to party!" help="Get notified when the party starts." v-model="checked" />`
-const checkboxProps = [
-  { name: "emphasis", required: false, type: "boolean" },
-  { name: "label", required: false, type: "string" },
-  { name: "modelValue", required: true, type: "boolean" },
-]
-const dateRange = ref({ maxDate: 0, minDate: 0 })
-const dateRangePickerCopy = `<DateRangePicker v-model="dateRange" />`
-const dateRangePickerProps = [
-  { name: "maxRange", required: false, type: "number" },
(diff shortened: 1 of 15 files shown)


dev/content/Forms.vue
```diff
@@ -190,6 +199,7 @@ const toggleProps = [{ name: "modelValue", required: true, type: "string" }]
           <div class="mt-1">
             <form @submit.prevent>
               <BaseInput
+                v-model="inputVals['baseInput']"
```
steven-klein, line 202: This highlights the primary value of allowing undefined v-model values.  Rather than needing to explicitly type your payload model, you should be able to do this instead:

```vue
<script setup lang="ts">
// won't require you to set every key, but will only allow keys currently on User 
const formVals = ref<Partial<User>>({})
</script>

<template>
    <BaseInput v-model="formVals['email']" type="email" label="Email" required />
</template>
```

It will be a bit experimental to see whether this is a good thing, but it's ultimately opt in.

src/composables/forms.ts
```diff
@@ -1,6 +1,79 @@
...
+export interface InputOption {
+  disabled?: boolean
+  help?: string
+  label: string
+  sublabel?: string
+  value: string | number
+}
+
+export interface BooleanInput extends Input {
+  modelValue?: boolean | null
```
steven-klein, line 21: This used to bite us all the time - now we have `null` support almost across the board as we continue to work with various `Null$t{ype}`'s

profsmallpine (approved):
🌟

PR #50, opened 2023-10-10
3. add batteries to inputs, improve error message display

steven-klein:
> PR 3 of 3 related to [Form Input Updates](https://www.notion.so/xylabs/8f16954aa3494675837f0cceb6a21c95?v=9d015d84f6fa48fe837c8ad0f6ab5be8&p=ecae8501d1924fa88b6561c05235759b&pm=s)

# What this does

- adds number parsing to `BaseInput type="number"` (no longer need to `v-model.number` to get a type number - a copy/paste of the default number directive in vue/core)
- adds default pattern attribute to `BaseInput type="email"` (used what exists in Portal currently)
- adds default pattern attribute to `BaseInput type="tel"` (used what exists in Portal currently)
- ~~adds default pattern attribute to `BaseInput type="password"` (used what exists in Portal currently)~~
- adds default red `*` to input label when required attribute is present (strips existing `*`)
- adds `useModel` composable to enable local input state for v-model and error props (this is necessary to keep the input state value on the `HTMLInputElement` when redraws occur and v-model is undefined - this is close to a copy/paste of the `useModel` composable that is currently being tested in vuejs/core)
- adds error message handling and display of default HTML validation error messages.
- adds min/max support to multicheckbox
- resolves DateRangePicker not responding to required attribute

## Notes

The general error handling and display works pretty well from my tests and continues to stay as close to default HTML form validation as possible, mostly reacting to `oninvalid` events.  Once you start down the path of setting error messages, you quickly find that the browser's default handling of error display is pretty sad, and relying on it is not a great experience particularly in longer forms.

There is a test form in the docs that will give you a quick sense of that throwing up required fields might look like with a custom error thrown on Multicheckboxes, that custom error is a wip and is not necessarily the best example of how a custom validation message might ultimately, but in general, the goal is that you just set the v-model:error prop.  The vast majority of our use cases for custom error messages is going to come from static forms like the login form, and isn't likely to have a heavy reactive use cases - which is where some compromises might come in the end.

### TODO's:

- [x] resolve whether there is a need to track the input element inside useInputField (this helps ensure default HTML validation continues to set focus state etc. when invalid fires)
- [x] the required attribute on DateRangePicker does not react to invalid events (this looks like it might be an old problem)
- [x] multicheckbox should never set `required` on the inputs (it will just make every checkbox required), but supporting a min/max like validation out of the box would likely relieve any need for custom validation and could support a `*` on the label.

Screenshot 2023-10-10 at 5 55 20 PM

Screenshot 2023-10-10 at 5 55 34 PM


steven-klein:
@DavidLarsKetch @profsmallpine this should be in a good final state.  As discussed, I've stripped out the error prop, cleaned up error label usage, and updated the lib exports.

dev/content/Forms.vue
```diff
@@ -189,39 +188,19 @@ const toggleProps = [
             class="xy-link"
             target="_blank"
             >input variations</a
-          >. Errors can be styled with <code>.xy-input-error</code>.
+          >. Errors can be set with the error prop.
```
profsmallpine, line 191: Didn't we drop this prop?

profsmallpine (approved):

dev/content/Forms.vue
steven-klein: 🤦

PR #54, opened 2024-02-09
composables for tab history and search params history

steven-klein:
# What this does

- adds `useTabHistory` for supporting tab navigation state via the browser query string: `?tab=my-tab`
- adds `useUrlSearchParams` for supporting complex search params state via the browser query string: `?q=find+me+something&q=also+this&isActive=true`

## Notes:

This is aimed at supporting [#1091](https://github.com/xy-planning-network/college-try/pull/1091) on Portal and establishing a common path forward for existing browser history/state support via query strings.

I ultimately landed on using a query param with `useTabHistory` since it keeps the `#` fragment free for anchors.  i.e. `/setting?tab=advisor-profile#fees`.

### DIFF
--- a/dev/content/Tools/UseUrlSearchParams.vue
+++ b/dev/content/Tools/UseUrlSearchParams.vue
@@ -4,16 +4,20 @@ import CodeSample from "../../helpers/CodeSample.vue"
 import ProseBase from "../../helpers/ProseBase.vue"
 
 interface SearchParams {
-    q: string
-    isActive: boolean
+  q: string
+  isActive: boolean
+  attributes: string[]
 }
  
 // NOTE: this should tyically come from the server in the form
 // of a page prop.  Though some exceptions may exist.
 const initialParams: SearchParams = {
-    q: "",
-    isActive: false
+  q: "",
+  isActive: false,
+  attributes: []
 }
+
+const attrOpts = [{label: "Type One", value: "type-one"}, {label: "Type Two", value: "type-two"}, {label: "Type Three", value: "type-three"}]
  
 const searchParams = useUrlSearchParams(initialParams)
  
@@ -29,12 +33,14 @@ const searchParams = useUrlSearchParams(initialParams)
     <ProseBase>
       <h4>Example Usage:</h4>
       <h5>Script Setup</h5>
+      <!-- prettier-ignore -->
       <CodeSample>{{
         `
-<script setup lang="ts">
+<script setup lang="ts"&gt;
 interface SearchParams {
    q: string
    isActive: boolean
+   attributes: string[]
 }
 
 // If you're attempting to hydrate fields with a request that contains search params.
@@ -49,11 +55,13 @@ const searchParams = useUrlSearchParams(props.initialParam)
       }}</CodeSample>
 
       <h5>Template</h5>
+      <!-- prettier-ignore -->
       <CodeSample language="html">{{
         `
-<template>
+<template&gt;
     <BaseInput v-model="searchParams.q" label="Search" type="search" />
     <Checkbox v-model="searchParams.isActive" label="Is Active" />
+    <MultiCheckboxes v-model="searchParams.attributes" label="Attributes" :options="attrOpts" />
 </template&gt;
         `
       }}</CodeSample>
@@ -63,6 +71,7 @@ const searchParams = useUrlSearchParams(props.initialParam)
         <h5>Use the input controls to mutate the search params ref and check the browser url for updates to the query string.</h5>
         <BaseInput v-model="searchParams.q" label="Search" type="search" />
         <Checkbox v-model="searchParams.isActive" label="Is Active" />
+        <MultiCheckboxes v-model="searchParams.attributes" label="Attributes" :options="attrOpts" />
     </div>
   </ComponentLayout>
 </template>
--- a/dev/content/Tools/UseUrlSearchParams.vue
+++ b/dev/content/Tools/UseUrlSearchParams.vue
@@ -1,0 +1,68 @@
+<script setup lang="ts">
+import { useUrlSearchParams } from "@/composables";
+import CodeSample from "../../helpers/CodeSample.vue"
+import ProseBase from "../../helpers/ProseBase.vue"
+
+interface SearchParams {
+    q: string
+    isActive: boolean
+}
+ 
+// NOTE: this should tyically come from the server in the form
+// of a page prop.  Though some exceptions may exist.
+const initialParams: SearchParams = {
+    q: "",
+    isActive: false
+}
+ 
+const searchParams = useUrlSearchParams(initialParams)
+ 
+</script>
+
+<template>
+  <ComponentLayout :show-badge="false" title="useUrlSearchParams">
+    <template #description>
+      useUrlSearchParams is a composable methods that keeps the values of a
+      record of params updates in the browser url.
+    </template>
+
+    <ProseBase>
+      <h4>Example Usage:</h4>
+      <h5>Script Setup</h5>
+      <CodeSample>{{
+        `
+<script setup lang="ts">
+interface SearchParams {
+   q: string
+   isActive: boolean
+}
+
+// If you're attempting to hydrate fields with a request that contains search params.
+// Parse the params on the server and send them as a page prop to your search component.
+const props = defineProps<{
+    initialParams: SearchParams
+}>
+
+const searchParams = useUrlSearchParams(props.initialParam) 
+</script&gt;
+        `
+      }}</CodeSample>
+
+      <h5>Template</h5>
+      <CodeSample language="html">{{
+        `
+<template>
+    <BaseInput v-model="searchParams.q" label="Search" type="search" />
+    <Checkbox v-model="searchParams.isActive" label="Is Active" />
+</template&gt;
+        `
+      }}</CodeSample>
+    </ProseBase>
+
+    <div class="space-y-5">
+        <h5>Use the input controls to mutate the search params ref and check the browser url for updates to the query string.</h5>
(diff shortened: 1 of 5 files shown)


src/composables/nav.ts
```diff
@@ -21,3 +29,204 @@ export interface ActionItem {
...
+
+    window.history.replaceState(
+      { tab: tab },
+      document.title,
+      `${window.location.pathname}${queryString}`
+    )
+  })
+
+  return {
+    activeTab,
```
profsmallpine, line 109: I expected to see an exported func to check if the tab is active. I am not doing anything more in archive than checking `isActiveTab`. Worth exporting the function to avoid `activeTab == 'tab-1'` as a 3rd export?

src/composables/nav.ts
```diff
@@ -21,3 +29,204 @@ export interface ActionItem {
...
+      label: string
+      value: string
+    }[]
+  >,
+  opts?: UseTabHistoryOpts
+) => {
+  const tabs = ref(initialTabs)
+  const config = {
+    initial: tabs?.value[0]?.value || "",
+    paramName: "tab",
```
profsmallpine, line 77: Since `opts.paramName` never overrides the config here, should it get passed at all? Or a different type for opts vs config? I expected `paramName: opts!.paramName ? opts!.paramName : 'tab'` here.

src/composables/nav.ts
```diff
@@ -21,3 +29,204 @@ export interface ActionItem {
...
+  watch(
+    searchParams,
+    (p) => {
+      const params = new URLSearchParams(window.location.search)
+
+      for (const key in p) {
+        const val = p[key]
+        params.delete(key)
+
+        if (val == "" || val == false || val == null || val == undefined) {
```
profsmallpine, line 210: Probably just not grokking the overall context, but I'm confused why false values are thrown away. I imagine `showEmails: false` and `showEmails: true` and an empty value trigger 3 separate queries. Getting rid of the key would block necessary behavior in my mind.

dev/content/Tools/UseUrlSearchParams.vue
```diff
@@ -0,0 +1,68 @@
...
+      }}</CodeSample>
+    </ProseBase>
+
+    <div class="space-y-5">
+        <h5>Use the input controls to mutate the search params ref and check the browser url for updates to the query string.</h5>
+        <BaseInput v-model="searchParams.q" label="Search" type="search" />
+        <Checkbox v-model="searchParams.isActive" label="Is Active" />
+    </div>
+  </ComponentLayout>
+</template>
```
profsmallpine, line 68: I find these examples really handy ✨

profsmallpine (changes requested):
Nice to get this moved into trails, couples q's here, but just minor stuffz

src/composables/nav.ts
steven-klein: Yep, not opposed to this - even had a moment where it was included, but dropped it.  I didn't see it at first, but I suppose `v-if="isActiveTab('tab-1')"` in a template is a bit easier on the eye's than `v-if="activeTab === 'tab-1'"`.

Is that what you're looking for here?

I couldn't quite get it nailed down, but there's a future where an `isActiveTab('my-tab')` method can also be type safe on `tabs[0].value` values, but all implementations led to the use of `as const` which created more headaches, and it ultimately didn't seem helpful enough to pursue.

profsmallpine: Yeah exactly `isActiveTab('my-tab')`. It's a small win, but I find myself writing it in each component for the ergonomics

src/composables/nav.ts
profsmallpine: Oh duh, my bad. My original question that I deleted was asking if `paramName` ever gets overridden. Why not just use tab and not make it configurable?

src/composables/nav.ts
steven-klein: > Getting rid of the key would block necessary behavior in my mind.

I probably overlooked some use cases in order to keep this first iteration simple (i.e. no configuration options).  On community search, this is what we're doing via the `removeNullishValues` and `removeFalsyValues` options in the `VueUse` version of [useUrlSearchParams](https://vueuse.org/core/useUrlSearchParams/?foo=bar&vueuse=awesome#type-declarations).

By itself,  `useUrlSearchParams` won't trigger queries as it's a one-way binding between some object with key/value pairs and `window.location.search` where `window.location.search` is mutated as the object is mutated, so generally, it's just reacting to input controls.  Queries are fired off based on mutations of the same object, but some client-side method and Axios ultimately determine the shape of that payload.

Axios, currently drops `nullish` values from payloads (as we have it configured), so I expect, we'll never need to record nullish values via `useUrlSearchParams`, primarily because our input controls which can be initialized with a null value, don't _typically_ allow the user to return to a null value once they've interacted with it (think radios and checkboxes) - though exceptions exist.  I believe a `SearchParams` like struct could also end up with a literal "null" string value if not carefully considered.

However, I can see how a `showEmails: null | true | false` might evaluate to 3 different results in a query, so if we supported specifically "false" values we should be able to cover those cases, I'm just not clear on how common that really is or whether we should do it by default at the moment.

There's also a little bit of nuance here around whether `useUrlSearchParams` should immediately evaluate initialParams or not.  Currently, it does not, since I wanted to avoid mutating browser history on initialization.

src/composables/nav.ts
steven-klein: It allows a page to potentially support two sets of tabbed navigation, as well as offering a bailout for some kind of unexpected naming conflict between another component that's mutating query params.  Both could be premature, supporting it just seemed trivial.

steven-klein: I think nested tabs, is most likely case?  But potentially a bad UI as well!

src/composables/nav.ts
steven-klein: @profsmallpine should we go one more step and updated this composable to `useTabNav(tabs, config)`, where `useHistory` is a config option?  That would expose `isActiveTab` in situations where having a tab history isn't desirable.

profsmallpine: At this point, I would want this to push to the browser's history in all use cases and I can't picture a use case where we would want to exclude that as a rule. I'm sure it exists, but I would probably start without so it's intentional. Not sure on the balance of predicting the future so that minor changes to usage forces a trees bump, but as you know, I default to less functionality at the start.

src/composables/nav.ts
```diff
@@ -21,3 +30,250 @@ export interface ActionItem {
...
+/**
+ * useTabHistory
+ *
+ * Example Usage:
+ *
+ * const {activeTab, isActiveTab, tabs} = useTabHistory([{label: "Tab One", value: "tab-1"}, {label: "Tab Two", value: "tab-2"}]})
+ *
+ * <Tabs v-model="activeTab" :tabs="tabs" />
+ *
+ * <div v-if="isActiveTab('tab-1')">Tab 1 Content</div>
```
profsmallpine, line 90: ✨

profsmallpine (approved):

src/composables/nav.ts
steven-klein: @profsmallpine let me know if you checked out the side effect of the change on the community search page in [#1091](https://github.com/xy-planning-network/college-try/pull/1091).  It does set all params on first mutation, i.e.

`/community/profiles?q=test+search&sort=&sortDir=&ethnicity=&feeFlat=false&feeHourly=false&feeMonthly=false&feeQuarterly=false&feeAUM=false&feeAdditional=false&genderIdentity=&hasEmploymentInterest=false&hasHiringInterest=false&hasPurchaseInterest=false&hasSaleInterest=false&hasMergeInterest=false&hasMenteeInterest=false&hasMentorInterest=false&persona=&teamSize=&state=&location=`

It should be fine, just a notable difference when there are many search params (which does seem like an outlier).

If you have a place in archive you want to pull this into before I merge and publish to NPM let me know.  Latest is available via `npm install @xy-planning-network/trees@dev`.

profsmallpine: Yeah that gets busy. Do we want to continue dropping null values and just set true/false? I was most concerned about the need to have false values in the param, less so with null and that keeps the url more manageable. Maybe it doesn't matter much as users don't give it a ton of attention.

steven-klein: > Do we want to continue dropping null values 

We are dropping them in the current implementation - we have to to support the parser/decoder on the server side.  False is being set in community search since most of the checkbox like filters are zero-valued to `false`.

> Maybe it doesn't matter much as users don't give it a ton of attention.

This is my expectation, but there is something that doesn't feel right about it.  Maybe we're just over thinking it.

I can't easily determine what key's have/haven't mutated with a watcher on an object. Typically, the callback receives `newVal`, `oldVal` arguments, but in the case of an object like this, Vue doesn't hold the `oldVal`.  Since there are array's in the mix I also can't just use a simple shallow copy, though I could do a deep copy, and in theory have a reference to work with.

steven-klein: or, we can relent and just make it an option `{removeFalsey: bool}`

src/composables/nav.ts
```diff
@@ -21,3 +30,257 @@ export interface ActionItem {
...
+            params.append(key, val.toString())
+          })
+          continue
+        }
+
+        // avoid setting falsey (zero) values until a truthy value was previously set
+        // otherwise all zero-valued keys will be populated with every mutation
+        if (!params.has(key) && !val) {
+          continue
+        }
```
steven-klein, line 268: @profsmallpine check out this latest little tweak.  I think this strikes the balance we are looking for.

src/composables/nav.ts
profsmallpine: Did some testing on portal, this strikes a good balance 👍 🚀

PR #60, opened 2024-04-12
add datetime-local to RFC 3339 handling in BaseInput

steven-klein:
# What this does

- handles read/write v-model transformation when BaseInput is type=datetime-local

## Note

vue@3.4 ships with custom v-model modifiers and a `defineModel` macro, which might be a cleaner long term path, but I don't expect many more of these kinds of two-way transforms being necessary, so I'd rather ship something out now, and look more closely at defineModel + useModal before reaching for them.

### DIFF
--- a/src/composables/forms.ts
+++ b/src/composables/forms.ts
@@ -243,3 +243,20 @@ export const looseToNumber = (val: any): any => {
   const n = parseFloat(val)
   return isNaN(n) ? val : n
 }
+
+/**
+ * converts an RFC 3339 string to a datetime-local input value string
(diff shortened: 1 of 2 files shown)


steven-klein:
@DavidLarsKetch I believe this will handle `datetime-local` inputs.  `npm install @xy-planning-network/trees@dev` if you want to do a quick test.

DavidLarsKetch (approved):
Haven't been able to force any bugs, so the transforming is working well. If we opt to not go with a custom component then 👍 on merging this PR.

PR #61, opened 2024-04-16
Spk/datetime input

steven-klein:
# What this does

- adds `DateTime` input for handling v-model mutations of date strings in RFC 3339 format
- exports `textInputTypes` to remove small TODO from FieldsSchema work in portal

## Notes

See [#1220](https://github.com/xy-planning-network/college-try/pull/1220) for the bump in portal with a small tweak to FieldsSchema.

### DIFF
--- a/src/composables/forms.ts
+++ b/src/composables/forms.ts
@@ -31,6 +31,10 @@ export interface DateRangeInput extends Input {
   startDate?: number
 }
 
+export interface DateTimeInput extends Input {
+  modelValue?: string | null
+}
+
 export interface TextLikeInput extends Input {
   modelValue?: string | number | null
   type: TextInputType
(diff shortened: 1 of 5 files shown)


src/composables/forms.ts
```diff
@@ -63,19 +67,21 @@ export const defaultInputProps = {
...
+  "password",
+  "search",
+  "tel",
+  "text",
+  "time",
+  "url",
+  "week",
+] as const
+
+export type TextInputType = (typeof textInputTypes)[number]
```
steven-klein, line 84: This will allow me to remove a small copy/paste in the `FieldsSchema` work on portal, see [#1220](https://github.com/xy-planning-network/college-try/pull/1220)

src/lib-components/forms/DateTimeInput.vue
```diff
@@ -0,0 +1,73 @@
+<script setup lang="ts">
```
steven-klein, line 1: This is effectively a copy/paste of `BaseInput` that drops the `type` prop and any other special type specific handling.  I'd rather not bother with FlatPickr integration just yet, let's talk about whether `DateRangePicker` should adopt a similar date string handling and bite them off at the same time (potentially).

src/lib-components/index.ts
```diff
@@ -109,6 +111,7 @@ export interface TreesComponents {
   BaseInput: typeof BaseInput
   Checkbox: typeof Checkbox
   DateRangePicker: typeof DateRangePicker
+  DateTime: typeof DateTime
```
steven-klein, line 114: This is the place to nit the component name.  `DateTime`, DateTimeInput`, `DateTimeLocal` all seemed reasonable to me.  I went with the shortest after finding out that the `<input type="datetime" />` in the HTML spec is obsolete.

profsmallpine (approved):
`DateTime` sounds good to me

DavidLarsKetch (approved):
🪨 solid

PR #63, opened 2024-04-22
Style cleanup on tables and <ActionsDropdown />

DavidLarsKetch:
On `<DynamicTable />`...
- A column header title & sort button may not flow nicely when the title is longer. So, wrap it in a `flex` container
- Adds `bg-gray-50` on even rows to help visually differentiate
    - Also removes `bg-white` from the `<ActionsDropdown />` menu icon since that may or may not mesh with the row's background color
- Fixes the default "No items found..." row length to fill in when overflowing

_Before_
Screenshot 2024-04-22 at 12 20 41 PM
Screenshot 2024-04-22 at 12 20 47 PM

_After_

https://github.com/xy-planning-network/trees/assets/27823069/da0d5854-1fe3-4278-a932-898c9972bf76

Screenshot 2024-04-23 at 11 02 49 AM

### DIFF
--- a/src/lib-components/lists/DynamicTable.vue
+++ b/src/lib-components/lists/DynamicTable.vue
@@ -304,7 +304,7 @@ loadAndRender()
           </tr>
         </thead>
 
-        <tbody class="bg-white divide-y divide-gray-200">
+        <tbody class="bg-white">
           <tr
             v-for="(row, rowIdx) in rows"
             :key="rowIdx"
--- a/src/lib-components/lists/DynamicTable.vue
+++ b/src/lib-components/lists/DynamicTable.vue
@@ -244,54 +244,56 @@ loadAndRender()
               class="px-6 py-3 text-xs font-medium tracking-wider text-gray-900 uppercase bg-gray-50 leading-4"
               :class="col.alignment"
             >
-              <span v-if="col.title">{{ col.title }}</span>
-              <span
-                v-if="col.sort"
-                class="cursor-pointer"
-                @click.prevent="handleSort(col.sort as string)"
-              >
-                <svg
-                  v-if="currentSort !== col.sort"
-                  xmlns="http://www.w3.org/2000/svg"
-                  fill="none"
-                  viewBox="0 0 24 24"
-                  stroke="currentColor"
-                  class="h-5 inline"
+              <div class="flex items-center gap-2">
+                <span v-if="col.title">{{ col.title }}</span>
+                <span
+                  v-if="col.sort"
+                  class="cursor-pointer"
+                  @click.prevent="handleSort(col.sort as string)"
                 >
-                  <path
-                    stroke-linecap="round"
-                    stroke-linejoin="round"
-                    stroke-width="2"
-                    d="M8 9l4-4 4 4m0 6l-4 4-4-4"
-                  />
-                </svg>
-                <svg
-                  v-else-if="currentSortDirection == 'desc'"
-                  xmlns="http://www.w3.org/2000/svg"
-                  viewBox="0 0 20 20"
-                  fill="currentColor"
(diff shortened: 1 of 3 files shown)


src/lib-components/navigation/ActionsDropdown.vue
```diff
@@ -29,7 +29,7 @@ const { floatingStyles } = useFloating(trigger, wrapper, {
   <Menu as="div" class="relative flex justify-end items-center">
     <MenuButton
       ref="trigger"
-      class="w-8 h-8 bg-white inline-flex items-center justify-center text-gray-700 rounded-full hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-blue-500 disabled:opacity-50 disabled:cursor-not-allowed"
+      class="w-8 h-8 inline-flex items-center justify-center text-gray-700 rounded-full hover:text-gray-900 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-blue-500 disabled:opacity-50 disabled:cursor-not-allowed"
```
steven-klein, line 32: Just a note:  this should be OK.  The only other places I recall using `ActionsDropdown` is in `DescriptionLists` and on the mastermind group show page.  Letting the parent background through, is probably the most desirable style in most use cases.

src/lib-components/lists/DynamicTable.vue
```diff
@@ -306,6 +308,7 @@ loadAndRender()
           <tr
             v-for="(row, rowIdx) in rows"
             :key="rowIdx"
+            class="even:bg-gray-50"
```
steven-klein, line 311: I'm good with making this our default, though I subjectively think we could drop the border between rows when they are striped.  The extra border adds to the visual competition.  Archive uses tables quite a bit more than Portal, so @profsmallpine should probably weight in on moving all in on striped tables.

I'd expect we'd move the `DataTable` to the same style as well for consistency.

steven-klein (changes requested):

src/lib-components/lists/DynamicTable.vue
```diff
@@ -244,54 +244,56 @@ loadAndRender()
...
-                  xmlns="http://www.w3.org/2000/svg"
-                  viewBox="0 0 20 20"
-                  fill="currentColor"
-                  class="h-5 inline"
+              <div class="flex items-center gap-2">
+                <span v-if="col.title">{{ col.title }}</span>
+                <span
+                  v-if="col.sort"
+                  class="cursor-pointer"
+                  @click.prevent="handleSort(col.sort as string)"
```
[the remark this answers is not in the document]
DavidLarsKetch: I did find the target a bit small, personally, so I'll take a look.

src/lib-components/lists/DynamicTable.vue
profsmallpine: I don't find one as more compelling than the other, but I think it is best to pick one, either using border or stripes. The downside of stripes for me is that it makes even rows the same background color as the header which I find visually noisy. Typically, the headers will loose their background shading so that you put the visual emphasis on the differentiation of rows.

DavidLarsKetch: How do we feel about this option?

Striped, no border between rows, header to `bg-gray-100` from `bg-gray-50`.
Screenshot 2024-04-23 at 11 02 49 AM

steven-klein: Yep, should be OK.  We use `bg-gray-100` as the app shell background color, so outside a card wrapper it _might_ be a touch off, but not to a degree I think we need to sweat right now.  I think Portal currently wraps all tables in some kind of card like wrapper.

steven-klein (approved):

PR #64, opened 2024-05-14
Open up <DetailList /> to more configuration

DavidLarsKetch:
## Problem statement
`<DetailList />` is tightly coupled to its use cases in `second-child`. A lot is baked in by default and can't be configured, unlike how `<DataTable />` and `<DynamicTable />` are more flexible and allow for more use cases.

Recently, we identified `<DetailList />` as a candidate for replacing some custom markup, simplifying the page it appears on. But the default filtering, styling, and pagination made it a bad fit.

## What this does
This PR brings `<DetailList />` opens up configurability following `<DynamicTable />`'s pattern. Notably, it copies the `DynamicTableAPI`, allowing for the parent to call list resetting methods.

Next, this PR turns the date searching feature into an optional one. When turned off, the `<DetailList />` does not display with a shadowed border. The rationale there is the parent can supply this styling when desired. When the only markup in `<DetailList />` is the items themselves, the border shadowing is difficult to work with.

This PR also allows for overriding a few default params passed to the request that retrieves the list items. Sorting and the amount of results to retrieve per page can be overwritten.

Lastly, this PR adds a feature to turn off pagination navigation altogether. In conjunction with the new `perPage` field, this enables `<DetailList />` to fit into tighter spaces, such as on `college-try`'s dashboard `<NotifcationCard />`.

## TODO
- [ ] ~~Move over an `<EmptyState />` type of component for display instead of the `<slot name="empty" />` in `<DetailList />`?~~
- [x] Finalize open TODOs
- [x] Finalize `<DateFilter />` given the removal of `title`
- [x] Finalize decision on use of option interfaces in `<DynamicTable />`
- [x] Update `/dev`
- [ ] ~~Review, clean up for `grid`-style alternative layout use case, like in `college-try`'s `/community/profiles`~~
- [x] Review, clean up for use cases in `second-child`

---
## Previous Discussion
Putting my cards on the table with regards to how I'm thinking about this:

In our backpack, we want a set of components to display a list in a particular layout. We want to support filtering the result set, sorting the result set, navigation through pages of results, and user actions able to be taken on a particular item. Many times, we want the components to fetch the data themselves and offload user interaction on the above functionality to the component. Less often, we supply the data to the component.

In that vein, a table is a kind of list. It's a layout style for a list. It seems `<DetailList />` is a columnar layout. There's mention of need for a grid layout (like for Community Search profiles). Ideally, one can swap between these 3 styles without substantial changes in configuration, potentially only dropping configuration that isn't supported in one layout. This way, ideating can move quickly, but more importantly, this approach simplifies the mental model required to interact with them.

To get there, I believe we need a shared configuration interface between `<DataTable />`, `<DynamicTable />` and `<DetailList />`. The `List*Options` interfaces are the intended way of getting there. Each component composes together the set of options that are applicable to its currently supported functionality. In many cases, their fields are optional and the component brings its own defaults.

`<DetailList />` isn't quite in alignment with `<DynamicTable />`, yet. I'd love to see the former remove the need for `<slot>`s altogether. We'd get there by adopting the `TableColumns` pattern, it seems. Then, we'd also need an idea of `TableActions`, since `second-child`'s templates have actions. These tasks seem like good candidates for the next pass, to me.

### DIFF
--- a/src/composables/list.ts
+++ b/src/composables/list.ts
@@ -2,21 +2,34 @@ import { DateRangeProps } from "./date"
 
 export interface DetailListConfig {
   /**
-   * date turns on the date-range filter input
-   */
-  dateSearch?: boolean | DateRangeProps
-
-  defaultSort?: string
-  defaultSortDir?: SortDir
-
-  /**
    * alwaysHideNav disables navigation components from displaying whatsoever.
    * This makes the List as a "single-page" limiting the number of results via perPage.
    */
   alwaysHideNav?: boolean
 
+  /**
+   * date turns on the date-range filter input
+   */
+  dateSearch?: boolean | DateRangeProps
+
+  /**
+   * defaultSort overrides the default field to sort items on.
+   */
+  defaultSort?: string
+
+  /**
+   * defaultSortDir overrides the default order to sort items with.
+   */
+  defaultSortDir?: SortDir
+
+  /**
+   * perPage overrides the default number of items per page to retrieve.
+   */
   perPage?: number
 
+  /**
+   * url is the endpoint to retrieve list data from
+   */
   url: string
 }
 
--- a/src/composables/list.ts
+++ b/src/composables/list.ts
@@ -1,38 +1,3 @@
-import { DateRangeProps } from "./date"
-
-export interface DetailListConfig {
-  /**
-   * alwaysHideNav disables navigation components from displaying whatsoever.
-   * This makes the List as a "single-page" limiting the number of results via perPage.
-   */
-  alwaysHideNav?: boolean
-
-  /**
-   * date turns on the date-range filter input
-   */
-  dateSearch?: boolean | DateRangeProps
-
-  /**
-   * defaultSort overrides the default field to sort items on.
-   */
-  defaultSort?: string
-
-  /**
-   * defaultSortDir overrides the default order to sort items with.
-   */
-  defaultSortDir?: SortDir
-
-  /**
-   * perPage overrides the default number of items per page to retrieve.
-   */
-  perPage?: number
-
-  /**
-   * url is the endpoint to retrieve list data from
-   */
-  url: string
-}
-
 export interface DetailListAPI {
   /**
    * Force refresh the table data with the current api params state
--- a/src/composables/list.ts
+++ b/src/composables/list.ts
@@ -1,6 +1,6 @@
 export interface DetailListAPI {
   /**
-   * Force refresh the table data with the current api params state
+   * Force refresh the list data with the current api params state
    * @returns void
    */
   refresh: () => void
--- a/src/composables/list.ts
+++ b/src/composables/list.ts
@@ -1,0 +1,36 @@
+import { DateRangeProps } from "./date"
+
+export interface DetailListConfig {
+  /**
+   * date turns on the date-range filter input
+   */
+  dateSearch?: boolean | DateRangeProps
+
+  defaultSort?: string
+  defaultSortDir?: SortDir
+
+  /**
+   * alwaysHideNav disables navigation components from displaying whatsoever.
+   * This makes the List as a "single-page" limiting the number of results via perPage.
+   */
+  alwaysHideNav?: boolean
+
+  perPage?: number
+
+  url: string
+}
+
+export interface DetailListAPI {
+  /**
+   * Force refresh the table data with the current api params state
+   * @returns void
+   */
+  refresh: () => void
+  /**
+   * Reset the table data back to page 1 and load
+   * @returns void
+   */
+  reset: () => void
+}
+
+export type SortDir = "asc" | "desc"
--- a/src/composables/list.ts
+++ b/src/composables/list.ts
@@ -5,7 +5,7 @@ export interface DetailListAPI {
    */
   refresh: () => void
   /**
-   * Reset the table data back to page 1 and load
+   * Reset the list data back to page 1 and load
    * @returns void
    */
   reset: () => void
--- a/src/lib-components/layout/DateFilter.vue
+++ b/src/lib-components/layout/DateFilter.vue
@@ -1,27 +1,27 @@
 <script setup lang="ts">
-import { DateRange } from "@/composables/date"
 import { ref } from "vue"
+import { DateRange } from "@/composables/date"
+import { SortDir } from "@/composables/list"
 import DateRangePicker from "../forms/DateRangePicker.vue"
 import Select from "@/lib-components/forms/Select.vue"
 
 const props = defineProps<{
   dateRange: DateRange
-  sortDir: string
-  title: string
+  sortDir: SortDir
 }>()
 const dateRange = ref<DateRange>(props.dateRange)
-const sortDir = ref<string>(props.sortDir)
-const sortOptions = [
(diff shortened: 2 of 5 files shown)


src/lib-components/lists/DynamicTable.vue
```diff
@@ -95,22 +126,16 @@ const { columns, hasActions, isEmptyCellValue, rows } = useTable(
...
+const currentSortDirection = ref(props.config.sort?.defaultSortDir)
 const dateRange = ref({
   minDate: 0,
   maxDate: 0,
 })
 
 const pagination = ref({
   page: 1,
-  perPage: 10,
+  perPage: props.config.pagination?.perPage || 10,
```
DavidLarsKetch, line 138: @steven-klein example where the compiler yelled about me about `config.pagination` being potentially `undefined`, which I didn't expect. I end up having to set the default again, which maybe means we just don't use `withDefaults`?

src/composables/list.ts
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * alwaysHideNav disables navigation components from displaying whatsoever.
+   * This makes the List as a "single-page" limiting the number of results via perPage.
+   */
+  alwaysHideNav?: boolean
+
+  /**
+   * date turns on the date-range filter input
+   */
+  dateSearch?: boolean | DateRangeProps
```
steven-klein, line 13:
```suggestion
  disableDate?: boolean
```

Looks like this prop never makes use of `DateRangeProps`.  From the current and new use cases, we can probably punt on allowing it to be configurable and revisit as needed.  I'm unsure how to think about sane defaults on the component in general, but for consistency I suppose we work from the assumption you get the kitchen sink of tools and disable what you don't need.  So this could just be `disableDate?: boolean` that defaults to false.  It feels weird, but with boolean props I'm usually aiming to keep their default to false so that you can `<DetailList disable-date />`  as opposed to `<DetailList date-search="false" />`.

src/lib-components/layout/DateFilter.vue
```diff
@@ -1,27 +1,27 @@
...
 import { ref } from "vue"
+import { DateRange } from "@/composables/date"
+import { SortDir } from "@/composables/list"
 import DateRangePicker from "../forms/DateRangePicker.vue"
 import Select from "@/lib-components/forms/Select.vue"
 
 const props = defineProps<{
   dateRange: DateRange
-  sortDir: string
-  title: string
```
steven-klein, line 10: I'm okay with dropping this prop.  Looking at the Archive instance in the `ReviewsList` I don't know that it adds a ton of value, the context still appears clear without it.  @profsmallpine any concerns there?

src/lib-components/lists/DetailList.vue
```diff
@@ -1,45 +1,44 @@
...
-    url: string
-  }>(),
-  {
-    refreshTrigger: 0,
-    reloadTrigger: 0,
-  }
-)
+const props = defineProps<{
+  config: DetailListConfig
+}>()
```
steven-klein, line 11: I'd suggest we flatten these props.  Tables made use of generics so that we could gain some notion of type safety on the actions and columns configurations.  We don't have a need for that here currently, and instead it just means we'll find ourselves defining configuration props inside the script tag without any additional benefits.  I'd rather just put what will be mostly static configuration props right on the component in the template.

It's notable that `refreshTrigger` and `reloadTrigger` are dropped from support here, as Archive makes use of `reloadTrigger`.  I think that's okay, as `defineExpose` has us covered, and it would be nice to narrow our usage to one way of doing this kind of thing.  @profsmallpine any issues here?  You'd need to expose the `reload` method from `ReviewsList.vue` so your parent can call it.

src/lib-components/lists/DetailList.vue
```diff
@@ -1,45 +1,44 @@
...
   maxDate: 0,
 })
-const hasContent = ref(true)
+
 const items = ref<any[]>([])
+
 const pagination = ref({
   page: 1,
-  perPage: 10,
+  perPage: props.config.perPage || 10,
```
steven-klein, line 22: After flattening the props we can leverage `withDefaults` for setting these types of default values and limit the number of potentially `undefined` props.

src/lib-components/lists/DetailList.vue
```diff
@@ -48,47 +47,60 @@ const loadAndRender = (checkForContent: boolean): void => {
...
+        @date-range-changed="dateRange = $event"
+      />
+    </div>
 
-    <div v-if="hasContent" class="shadow overflow-hidden sm:rounded-md border">
+    <div
+      v-if="hasContent"
+      class="overflow-hidden"
+      :class="{
+        'shadow sm:rounded-md border': filtersAreConfigured,
```
steven-klein, line 101: We should make the shadow and border a configuration prop.  I'm trying to be less afraid of them, but it's a more stable and reliable option than letting callers pass class names into components.  We can do that in a pinch, but we shouldn't design components around that pattern.  We want to be able to make style changes without feeling like we have to go and examine every use case to see if there's conflicting `class` attribute being inherited.

I like the idea behind considering whether the existence of filters should drive the styling, but I'm less confident that will give us a meaningful runway.

`borderless?: bool` with a default of false seems like a sane place to start, I'd even apply it to the `li`, until we can narrow down how best to make this more configurable.

steven-klein (changes requested):
Overall, this looks good.  It's incremental and we can build on it.  I didn't really foresee the props being consolidated into a single property after the first draft.  I can see why it went that direction, but I think we should stay flat for now.

src/lib-components/layout/DateFilter.vue
profsmallpine: Yeah all good dropping this

src/lib-components/lists/DetailList.vue
profsmallpine: Yeah, that works

src/lib-components/lists/DetailList.vue
```diff
@@ -48,47 +93,58 @@ const loadAndRender = (checkForContent: boolean): void => {
...
-    // This lets parent components trigger a refresh of the current page depending on external actions.
-    pagination.value.page = 1
-    loadAndRender(true)
-  }
-)
+const showPaginator = computed(() => {
+  return !props.disableNavigation && hasContent
+})
+
+const reloadTable = (): void => {
```
steven-klein, line 115:
```suggestion
const reset = (): void => {
```

Since we don't call this in the component, `reset` seems okay to map it to the `DetailListAPI` interface.

src/lib-components/lists/DetailList.vue
```diff
@@ -48,47 +93,58 @@ const loadAndRender = (checkForContent: boolean): void => {
...
+  pagination.value.page = 1
+  loadAndRender()
+}
 
-loadAndRender(true)
+const publicMethods: DetailListAPI = {
+  refresh: loadAndRender,
+  reset: reloadTable,
```
steven-klein, line 122:
```suggestion
  reset: reset,
```

steven-klein (approved):
Looks like it's in good shape to get it released.  I made a handful of nit suggestions, one for the docs and to strip out references to the word "table" in the component.

I'll throw a dev release up on NPM in case it's useful for confirming the use case.

PR #67, opened 2024-07-31
update value handling in select component, apply trim to text and tex…

steven-klein:
# What this does

- refactors the onChange event in the `Select` component to use the value property on the prop options instead of using the value on the input (allows for number support in options)
- applies the trim method to Textarea and BaseInput to drop the need of v-model.trim modifiers

## Notes

After looking closely at the `HTMLSelectElement` spec, it appears there is a way to determine which option was selected via an index value that can be mapped back to the options set on the component props.  This allows us to read the exact value set on the option allowing for both string and number support, technically even mixed uses.  The only situation not supported is mixed types where the string value is the same, ex: '3' and 3.  This potentially could be handled as well if we handle setting the `selected` attribute ourselves, but it seems unlikely to be necessary and may even be better classified as something we shouldn't do at all.  If a valid cases for it arises, I've got some ideas though.

Looking through Portal and Archive the use of the `trim` modifier is pretty heavy.  We've had that baked into. `FormDisplay` since the start, so it seems like a very useful default handling for those input types.

### DIFF
--- a/dev/content/Forms.vue
+++ b/dev/content/Forms.vue
@@ -183,7 +183,10 @@ const toggleProps = [
             class="xy-link"
             target="_blank"
             >input variations</a
-          >.
+          >. When the type attibute is set to 'number' the value returned will
+          be a number making v-model.number modifiers unnecessary. Likewise, the
+          `trim()` method is applies to all string values making the
+          v-model.trim modifier unnecessary.
         </template>
 
         <div>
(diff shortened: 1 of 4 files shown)


profsmallpine (approved):
Cool solution for the select 🥂

steven-klein:
@profsmallpine @seedifferently I'll leave this open in case anyone wants a glance in the next day or two.  Feel free to merge it if you want to move forward with it on Archive.  See [#1388](https://github.com/xy-planning-network/college-try/pull/1388/files) for the tweaks to `FormDisplay`, though that update shouldn't be required to get the benefits here.

PR #70, opened 2024-08-24
docs tweaks

steven-klein:
# What this does

This is a WIP ugly update to how docs are written for Trees.  A number of pain points have made writing docs difficult, this starts to alleviate some of them.  Specifically, adding code examples has been frustrating with how Vue SFC's compile and how prettier parses those files.  This starts to introduce markdown files that for now are mostly used for displaying code samples.

- refactored page declarations, home page layout, and app entry point logic
- makes numerous file name and file organization changes (WIP - no concrete direction here, but it doesn't fit with app patterns)
- migrates old code samples into markdown files
- fixes lack of type checking on `/docs` directory via tsconfig updates and global component type declarations


DavidLarsKetch (commented):
Looks great, only styling tweak I'd look into is the font size for code samples. Given the narrow layout for a doc, I in Tools you end up with a lot of overflow to scroll through. There's some with the original samples, but a lot less has been chopped off such that you can kind of guess what the end of a line is.

dev/app.vue
```diff
@@ -0,0 +1,96 @@
...
+ * determine the app layout on load
+ */
+const appLayout = (() => {
+  const key = "trees-app-layout"
+  const layout = new URLSearchParams(window.location.search).get("layout")
+  if (layout) {
+    window.localStorage.setItem(key, layout)
+  }
+
+  return window.localStorage.getItem(key) || "SidebarLayout"
```
profsmallpine, line 16: I would use `StackedLayout` as the default here to maximize horizontal space.

dev/docs/StackingContextDocs.md
```diff
@@ -0,0 +1,35 @@
+#### Tailwind CSS z-index Values
+
+| Component                        | z-index |
+| -------------------------------- | ------- |
+| Popover/Tooltip/ActionsDropdown  | z-[5]   |
```
profsmallpine, line 5: 🌟

dev/docs/UseBaseAPIDocs.md
```diff
@@ -0,0 +1,110 @@
+#### The useBaseAPI Interface
+
+```ts
+/**
+ * UseBaseAPIOptions extends Trees/ReqOptions
+ * these options are used only in the instantiation
+ * of a new UseBaseAPI composable
+ */
+export interface UseBaseAPIOptions extends ReqOptions {
```
profsmallpine, line 9: I wonder if 1. we could write our implementation such that it's importable here rather than redefining and 2. if it's even possible to import code here from elsewhere in the project. Just food for thought, nothing to do now

dev/pages/Elements.vue
```diff
@@ -1,8 +1,8 @@
 <script setup lang="ts">
 import { ref } from "vue"
-import ColorRow from "../../dev/helpers/ColorRow.vue"
+import ColorRow from "../components/ColorRow.vue"
```
profsmallpine, line 3: It would be nice to remove `max-w-3xl` from all of our page containers to get more horizontal space.

profsmallpine (changes requested):
Defaulting to more horizontal space is the only change request here. Accomplished by removing `max-w-3xl` and defaulting to the top nav.

This is a great start, look forward to seeing how this develops. Would love to have this be part of my workflow whenever I start a new page and layout for a feature.

steven-klein:
@DavidLarsKetch @profsmallpine I'll add the requested changes on once I merge it all together.  The doc's width change will make a mess of the diffs.

PR #71, opened 2024-08-24
initial progress circle components

steven-klein:
# What this does

- adds two progress circle components for the incoming New App and Account Setup work
- adds a new docs page called "Indicators"

## Notes

There's currently no expectation that these progress indicators will function like nav items and tabs, so they're pretty dumb.  The current step is indicated by the "index" of the step in "steps", i.e. step 1 is current with an index of 0.  If this doesn't feel intuitive, let me know, and we can get it changed before the first use cases.

### DIFF
--- a/dev/pages/Indicators.vue
+++ b/dev/pages/Indicators.vue
@@ -1,0 +1,128 @@
+<script setup lang="ts">
+import { ref } from "vue"
+import User from "@/composables/user"
+
+defineProps<{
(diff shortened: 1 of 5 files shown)


DavidLarsKetch (approved):
All looks pretty self-explanatory to me!

profsmallpine (approved):
Indexing on 0 for the current step is more intuitive for me

PR #81, opened 2024-10-04
Fix: addresses `sr-only` component impacting layout

DavidLarsKetch:
## Problem statement
A `sr-only` or "hidden" component factors is not ignored by classes like `space-x/y-*`, etc. Here's an example where the `<input>` does not exist:
Screenshot 2024-10-04 at 9 32 11 AM
Here's where it does:
Screenshot 2024-10-04 at 9 27 23 AM

## What this does
This PR fixes the one case of this bug in `forms/MultiCheckboxes`. It solves it by moving the `<input />` to the end of the the `<fieldset />` that has a `space-y-6`. Here's what that looks like:
Screenshot 2024-10-04 at 9 28 36 AM

There appear to be no other components using `sr-only` where this layout bug can pop up.

If I had to explain why this is a bug - and why this PR fixes it - unfortunately, I come up a bit short.

This indicates it makes sense we're running into it: https://github.com/tailwindlabs/tailwindcss/issues/3857

But, that `<fieldset />` has a `relative` class and the `<input />` has position classes. So, my mental model suggests the `<input />`is outside the flow of its sibling nodes. The only explanation I can come up with is the "cascading" part of CSS - might be `space-x/y-*` is applied before `top-*` and `left-*`.

### DIFF
--- a/src/lib-components/forms/MultiCheckboxes.vue
+++ b/src/lib-components/forms/MultiCheckboxes.vue
@@ -112,17 +112,6 @@ const setValidationError = () => {
       <InputError :id="aria.errormessage" class="mt-1" :text="errorState" />
     </div>
 
-    <!--Hidden input for custom validation-->
-    <input
-      v-if="countError"
-      ref="errorInput"
-      required
-      class="sr-only top-1 left-1"
-      aria-hidden
-      type="checkbox"
-      @invalid="setValidationError"
-    />
-
     <div class="flex">
       <div
         class="grid gap-y-5"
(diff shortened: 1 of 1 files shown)


steven-klein (approved):
Damn, there's so much CSS specific nuance going on here, from margin collapse to CSS specificity and precedence. As Adam pointed out in the discussion, the spacing utility is working as expected.  

In a nutshell, the optional label of a `MultiCheckbox` is ideally the first-child of the fieldset, so that it does not receive a margin-top from the spacing utility.  Moving the custom validation input to the bottom allows either the label or the input set to become "first-child" and avoid the additional vertical spacing.

PR #83, opened 2024-10-28
bump vue@3.5, adopt subset of new features

steven-klein:
# What this does

- bumps the peer dependency of Vue.js from 3.3 to 3.5
- bumps required 3rd party dependencies `vite`, `vite-plugin-vue`, `vue-tsc`
- adopts `defineModel` anywhere `useModel` or `modelValue` props were in use
- drops custom implementation of `useModel` (`useModel` is stable in core - `defineModel` is recommended for SFC)
- adopts `useTemplateRef` on all existing template ref's
- adds/updates convenience `@close` event to `CotnentModal`, `Modal`, and `Slideover`
- adopts `useId` from Vue.js where appropriate
- adds `LogPlugin` to Dev docs for logging convenience in templates
- exports Nav types to resolve warning in applications during dev
- renames `Item` nav type to `NavItem`
- resolves `vite` deprecation notices

### BREAKING CHANGES

- `useModel` has been removed - favor `defineModel` in SFC or `import { useModel } from "vue"` in composables
- navigation type `Item` renamed to `NavItem` for clarity
- `Slideover` @close event no longer includes reference to `modelValue` - does not appear to be useful, it will always be false.
- input component `update:model-value` events have type inference on the value.

### Install Notes

This bump requires a few additional dependency bumps in the apps and a small tweak to vite.config

```
# bump trees and vue
$ npm install @xy-planning-network/trees@latest, vue@latest

# bump dev dependencies
$ npm install -D vue-tsc@latest @vitejs/plugin-vue@latest vite@latest

# update vite.config
$ mv vite.config.ts vite.config.mts
```

### vue@3.4 and vue@3.5 TLDR;

A number of reactivity optimizations have been made alongside the new API features that shipped in vue@3.4 and vue@3.5.  These are the most significant changes to check out:

- [More Efficient Reactivity System](https://blog.vuejs.org/posts/vue-3-4#more-efficient-reactivity-system) - the primary concern here was that we may have unintentionally been relying on compute side effects to trigger callbacks.   I've found no negative side effects to this change so far.
- [defineModel is Now Stable](https://blog.vuejs.org/posts/vue-3-4#definemodel-is-now-stable) - `defineModel` is a compiler macro (think `defineProps`) that creates quite a bit of convenience for wiring up `v-model` in components.  Previously, we had a custom implementation of `useModel` to give us half of the value of `defineModel`.
- [useTemplateRef()](https://blog.vuejs.org/posts/vue-3-5#usetemplateref) - The main reason to adopt this is to leverage the automatic inference of the template type.  This appears to work quite well, though there are still edge cases where auto-inference fails, and the template type needs to be explicitly typed as before.
- [useId()](https://blog.vuejs.org/posts/vue-3-5#useid) - a convenience method for unique ID's - this composable still references `getCurrentInstance`, so doesn't completely replace our `Unique` utility, but get's very close.

### More will be revealed

Additional updates that are worth taking a peek at as we may want to consider adopting them in the future.

- [v-bind Same-name Shorthand](https://blog.vuejs.org/posts/vue-3-4#v-bind-same-name-shorthand)
- [Reactive Props Destructure](https://blog.vuejs.org/posts/vue-3-5#reactive-props-destructure) - there's some nuance with how destructured props are passed to composables that I need to spend a little more time with, but it looks promising and could help clean up `props.propName` calls.


src/lib-components/forms/DateTimeInput.vue
```diff
@@ -6,36 +6,27 @@ import {
...
 const props = withDefaults(defineProps<DateTimeInput>(), defaultInputProps)
+const modelState = defineModel<DateTimeInput["modelValue"]>({
+  ...defaultModelOpts,
+  get: (v) => {
+    return v ? toDatetimeLocal(v) : v
+  },
+  set: (v) => {
+    return v ? new Date(v).toISOString() : v
+  },
+})
```
steven-klein, line 26: This was one of the nicest instances where the `defineModel` features really cleaned things up.

src/lib-components/forms/BaseInput.vue
```diff
@@ -5,34 +5,46 @@ import InputError from "./InputError.vue"
...
+  //
+  // Example: calling v.trim() may prevent the use of the space character by the user.
+  //
+  // https://github.com/vuejs/core/commit/de174e1aa756508c7542605a448e55a373afb1ed#diff-f1e0971afb7e61e75d37d0fc7d7f993c9681dddb7c689cb455d217a246b3d6f0
+  set: (v) => {
+    if (props.type === "number") {
+      return looseToNumber(v)
+    }
 
-defineEmits(["update:modelValue", "update:error"])
```
steven-klein, line 21: `defineModel` handles declaring `update:modelValue` for you, so you can skip that step.  `"update:error"` is a crusty and is not an event we've been supporting, so you'll see all of them being removed from the input components.

src/lib-components/layout/DateFilter.vue
```diff
@@ -21,12 +21,16 @@ const emits = defineEmits<{
...
+    emits("sort-dir-changed", sortDir)
+  }
 }
 
-const dateRangeChanged = (dateRange: DateRange) => {
-  emits("date-range-changed", dateRange)
+const dateRangeChanged = (dateRange: DateRange | undefined) => {
+  if (dateRange) {
+    emits("date-range-changed", dateRange)
+  }
```
steven-klein, line 33: We didn't explicitly type the `update:model-value` event in our input components, so they've all been resolving to `any`.  `defineModel` is now inferring the type of the value in the callback based on the potential initial values, which will often include `null` and `undefined` depending on the input component.  A little type guarding will be necessary where we bind to `@update:model-value`.  I may be able to improve the type narrowing with generics in the future, but it's not terribly common at the moment.

src/lib-components/forms/BaseInput.vue
```diff
@@ -5,34 +5,46 @@ import InputError from "./InputError.vue"
...
-  modelState,
-  inputID,
-  isRequired,
-  onInvalid,
-  inputValidation,
-} = useInputField(props)
+    return v
+  },
+})
+const inputRef = useTemplateRef("input")
```
steven-klein, line 40: I ran into an edge case with `useTemplateRef` where using the same value for the variable name and the identifier threw a Vue.js warning.  i.e. `const input = useTemplateRef("input")`.

Guidance and style guide appears to suggest the `Ref` postfix naming convention, so I've updated our instances to follow that convention.

DavidLarsKetch (approved):

PR #88, opened 2025-03-18
[DEV-3410] Profsmallpine/max date on range

profsmallpine:


steven-klein (commented):
Looking at the prop's for this input, I'm wondering how `maxDate` and `maxRange` should relate to each other.  I would expect `maxRange` to consider my `maxDate` when used together.  Let me know if you have other intuitions there - it looks like in general we just ignore `maxDateOpt` and `startDate` when `maxRange` is used with tables/lists.

profsmallpine:
Yeah, it's a good point on maxDate. I would expect the maxDate (or startDate/minDate) to override what the range is trying to accomplish if it is beyond that day. So with a maxRange of 10 and a maxDate of March 20th, I would expect the allowable range to be March 18-20 when selecting today as opposed to what I assume it does now (override from March 18-28 given the range max prop). Let's sleep on it and chat tomorrow. Without an actual use case, not sure what will make sense down the road..

steven-klein (approved):
Looks good.  I took a quick pass at the logic for having min/max work with a range and had it working within reason, but UI was a little funky.   With a fresh pass when we don’t need immediate functionality I’m sure we’ll find a good landing spot.

PR #91, opened 2025-04-01
number inputs

steven-klein:
# What this does

- introduces a `NumberInput` component that masks all number inputs according to US-en number formatting with support for a `money` type to handle the conversion between a monetary string input to US cents as a number

### Notes

- Number handling in `BaseInput` has been good, but trying to support masking and number/money conversions for the backend is enough bloat that I'd prefer to separate them.  Longer term, it moves `BaseInput` away from supporting `number` types at all to limit type juggling on v-model values in a parent.
- Supporting a combination of prefix, precision > 0, and negative values creates some user input edge cases that can be tricky to handle.  ex: Typing a `.` character and not having it set in the input field is due to the interaction of trying to store a number type from an incomplete string input.  That's where the `masked` computed variable comes in.  It's more complicated than I'd like, but appears to work as intended.  An alternative approach is to lazily set the `modelState` on blur, but this had its own subtle headaches.
- We're no longer using the `step` attribute anywhere that is either necessary or that precision can't handle.   There's no clear way to support `step` with masking.  I suspect we don't really have a need for it long term - punting on worrying about it for now.
- The money mask configuration is intentionally opinionated at the moment based on its current use case.  If there's a need for something like a `dollars` input that is still storing data as `Money` let's deal with that if it shows up.
- `FieldSchema` will call `NumberInput` over `BaseInput` (only handful of instances between portal and archive) with a small tweak while I evaluate swapping out `BaseInput type="number"`

### Discussion

I'd like to call it out of scope for now so we can get money handled for team members, and I'm not looking to add disruption to Benchmarking, but rather than having prefix/postfix characters in the masked input I'd like to see us move toward a UI that bakes this prefix/postfix into the input where it is static and non-interactive.  It helps limit special handling on the input values on our side and avoids some odd user interactions when capturing inputs.

Screenshot 2025-03-31 at 4 48 48 PM

### DIFF
--- a/src/lib-components/forms/NumberInput.vue
+++ b/src/lib-components/forms/NumberInput.vue
@@ -14,6 +14,7 @@ import { maskitoTransform } from "@maskito/core"
 import {
   MaskitoNumberParams,
   maskitoNumberOptionsGenerator,
+  maskitoParseNumber,
 } from "@maskito/kit"
 
 defineOptions({
@@ -80,7 +81,7 @@ const masked = computed({
       return
     }
 
-    let number = parseFloat(v.toString().replace(/[^\d.-]/g, ""))
+    let number = maskitoParseNumber(v)
     if (isNaN(number)) {
       modelState.value = null
       return
--- a/src/lib-components/forms/NumberInput.vue
+++ b/src/lib-components/forms/NumberInput.vue
@@ -1,0 +1,165 @@
+<script setup lang="ts">
+import InputLabel from "./InputLabel.vue"
+import InputHelp from "./InputHelp.vue"
+import InputError from "./InputError.vue"
+import {
+  useInputField,
+  defaultInputProps,
+  defaultModelOpts,
+} from "@/composables/forms"
+import type { NumericInput } from "@/composables/forms"
+import { computed, useAttrs, useTemplateRef } from "vue"
+import { maskito as vMaskito } from "@maskito/vue"
+import { maskitoTransform } from "@maskito/core"
+import {
+  MaskitoNumberParams,
+  maskitoNumberOptionsGenerator,
+} from "@maskito/kit"
+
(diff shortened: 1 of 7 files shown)


profsmallpine (approved):
I hate to remove the step attribute for any power users we might have, but that # is probably so low that it's not worth it..

Looks good! I like how it handles the cents value in money when typing e.g. `123.45`, just what I would expect.

steven-klein:
> I hate to remove the step attribute for any power users we might have, but that # is probably so low that it's not worth it..

Agreed.  I do expect I'll tinker with adding some keydown events at some point to support something like it.  We could also keep BaseInput number around to support cases where it's actually really useful, like `min=1, max=12` number inputs where formatting won't come into play.

steven-klein:
- [ ] need to resolve this edge case.

Screenshot 2025-04-01 at 5 37 11 PM

1. paste `1212.70`
2. position cursor behind `7`
3. hit backspace to delete `7`
4. keydown `6`

src/lib-components/forms/NumberInput.vue
```diff
@@ -86,8 +86,14 @@ const masked = computed({
...
 
+    // NOTE(spk): There's no Integer type in JavaScript all numbers are stored as floating point values
+    // to avoid approximations from decimal values avoid calling parseFloat and use parseInt on only the dollar
+    // portion when converting from string to number.
     if (props.type === "money") {
-      number = number * 100
+      let parts = number.toString().split(".")
+      let dollars = parseInt(parts[0], 10)
+      let cents = parts.length > 1 ? parseInt(parts[1].padEnd(2, "0"), 10) : 0
+      number = dollars * 100 + cents
```
DavidLarsKetch, line 96: Oh, floats, you never cease to disappoint.

PR #92, opened 2025-04-01
[DEV-3404] init bulk actions support on dynamic table

steven-klein:
# What this does

- adds bulk action support on `DynamicTable`
- adds a `TablePaginator` component to replace the existing `Paginator` on tables to allow larger table views and aid in larger bulk selections
- adds `placement` support to `ActionsDropdown` - I did not end up using this, but left it for the future.

### What this doesn't do

I need to find a stopping point on this one since its usage is still pretty narrow and what's here should give us something to work with for the existing use cases and some new ones.

- ~~the bulk action callback does not include the full row data, it only includes the `ids`, persistent mode makes this just a hair more difficult to add, but it's a small add when there is a little more time~~
- a unique `id` key is assumed to be on the resource record - this can be flexible in the future if needed, but doesn't seem useful enough right now
- persistent mode has been removed for now as it seems more likely to be difficult to maintain than it is useful

### Notes

- checkout https://github.com/xy-planning-network/college-try/pull/1899 for a more functional demo
- I'm interested in alternative thoughts on the default `perPage` options in the `TablePaginator` - these seemed reasonable and common, but I did not give them a lot of thought
- The `keydown` handling on the `TablePaginator` was cheap, so I threw it in.  It's far from perfect with the debouncing, but seems like a nice, subtle feature and gives us something to look at for `NumerberInput`
- A dynamic table will show the bulk selection checkboxes anytime you have available bulk actions, i.e. you can compute show: false to disable bulk actions on shared table components.
- `v-model:selected` is available on the table to persist selections across `v-if` scenarios

### DIFF
--- a/src/composables/table.ts
+++ b/src/composables/table.ts
@@ -136,4 +136,5 @@ export interface TableColumn<T = TableRowData> {
 export type TableCellAlignment = "left" | "center" | "right"
 export type TableRowData = Record<string, any>
 export type TableColumns<T = TableRowData> = TableColumn<T>[]
+
 export type TableRowsData = TableRowData[]
--- a/src/composables/table.ts
+++ b/src/composables/table.ts
@@ -13,7 +13,7 @@ export interface DynamicTableOptions {
 }
 
 export interface DynamicTableAPI {
-  deselectAll: () => void
+  clearSelection: () => void
   /**
    * Force refresh the table data with the current api params state
    * @returns void
@@ -79,6 +79,8 @@ export interface TableBulkActions<T extends TableRowData = TableRowData> {
    * an array of TableActionItem definitions
    */
   actions: TableBulkActionItem<T>[]
+  persistent?: boolean // whether to persist the selections across pagination, searching, sorting, and filtering
+  // max?: number // TODO(spk): restrain users from creating server bombing selections?
   isSelectable?: (data: T) => boolean
 }
 
--- a/src/composables/table.ts
+++ b/src/composables/table.ts
@@ -1,4 +1,4 @@
-import { VNodeChild } from "vue"
+import { VNodeChild, type ComputedRef } from "vue"
 import { ActionItem } from "@/composables/nav"
 import { DateRangeProps } from "./date"
 
@@ -12,7 +12,12 @@ export interface DynamicTableOptions {
   url: string
 }
 
-export interface DynamicTableAPI {
+export interface DynamicTableAPI<T = TableRowData> {
+  /**
+   * Clear the currently selected rows when bulk selections are enabled
+   * This method is called when refresh and reset methods are called.
+   * @returns void
+   */
   clearSelection: () => void
   /**
    * Force refresh the table data with the current api params state
@@ -24,6 +29,10 @@ export interface DynamicTableAPI {
    * @returns void
    */
   reset: () => void
+  /**
+   * The selected data records when bulk actions are enabled on the table
+   */
+  selectedData: ComputedRef<T[]>
 }
 
 export interface TableActionItem<T = TableRowData> extends ActionItem {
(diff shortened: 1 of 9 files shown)


src/composables/table.ts
```diff
@@ -52,6 +53,27 @@ export interface TableActionItem<T = TableRowData> extends ActionItem {
...
+export interface TableBulkActionItem extends ActionItem {
+  /**
+   * Whether or not the bulk action item is enabled.  Disabled actions are
+   * visible in the UI, but do not trigger click events.
+   */
+  disabled?: boolean
+  /**
+   * The callback method triggered by the action item buttons click event.
+   * @param selected the array of selected rows by the primary key `id`
+   * @param _ NOTE(spk): Holding for T[]
```
profsmallpine, line 65: Dumb question: leaving this spot in case we sent back an array of objects?

src/composables/table.ts
```diff
@@ -63,6 +85,21 @@ export interface TableActions<T = TableRowData> {
...
+   */
+  actions: TableBulkActionItem[]
+  /**
+   * whether to persist the selections across pagination, searching, sorting, and filtering
+   */
+  persistent?: boolean
+  /**
+   * a function that determines if the row can be selected for bulk actions
+   */
+  isSelectable?: (data: T) => boolean
```
profsmallpine, line 100: Noice

src/lib-components/lists/DynamicTable.vue
```diff
@@ -145,6 +140,119 @@ const hasContent = computed((): boolean => {
   return rows.value.length ? true : false
 })
 
+const selectionsPersisted = computed(() => {
+  return props.tableBulkActions.persistent ?? false
```
profsmallpine, line 144: til: `??`

profsmallpine (approved):
Nice add to the table 🥳

src/composables/table.ts
steven-klein: Yeah, my hunch is that we will have quite a few instances where we'll want the full record at hand in the action, and it made sense to me for it to be the second argument.  Even just being able to trigger a flash that says "Successfully updated ${resouce.name}" vs "Successfully updated ${id}" will be handy.  I may try to squeeze this in yet.

src/lib-components/navigation/TablePaginator.vue
```diff
@@ -0,0 +1,146 @@
...
+const pageSelectOpts = computed(() => {
+  if (props.pageOptions) {
+    return props.pageOptions
+  }
+
+  return [
+    { label: "5 per page", value: 5 },
+    { label: "10 per page", value: 10 },
+    { label: "20 per page", value: 20 },
+    { label: "50 per page", value: 50 },
```
DavidLarsKetch, line 42: Looks like a good first pass. I'm failing to think of many situations where, say, 100 or even 50 results would be helpful. Whenever I've used those larger per page numbers in apps, my impression its something like (1) I don't know what to search for (e.g., unfamiliar with the data, search sucks, etc.) or (2) I want to see 1 or 2 fields on every line item (e.g. credit card statement).

src/lib-components/lists/DynamicTable.vue
```diff
@@ -303,6 +412,21 @@ loadAndRender()
               class="px-6 py-3 text-xs font-medium tracking-wider text-gray-900 uppercase leading-4"
             />
           </tr>
+
+          <tr v-if="hasBulkActions && selected.length > 0">
```
DavidLarsKetch, line 416: What if this is at the bottom of the table, i.e., `<tfoot/>`?

When I click that first checkbox, my cursor is not hovering the item I clicked on, so can't click -> move -> click as fluidly. Coming in late with the review, so take it on advisement, haven't tested it out.

DavidLarsKetch (approved):
Very, very great work!

PR #101, opened 2025-08-21
mount the date range picker flyout menu inside the wrapper

steven-klein:
# What this does

This change has the date pickers fly out menu element appended to our inputs wrapper to keep the fly out elements context within whatever other element the input is placed.  This should resolve issues where a date range picker placed in a modal or slideover is inaccessible as currently Flatpickr will append the fly out to the body and Headless UI's focus traps will prevent interaction with it.

I've added a `DateRangePicker` inside a `Slideover` in the docs for a quick test.

### DIFF
--- a/dev/pages/Forms.vue
+++ b/dev/pages/Forms.vue
@@ -7,6 +7,7 @@ import {
   numericInputTypes,
   textInputTypes,
 } from "@/composables/forms"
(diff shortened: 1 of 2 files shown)


DavidLarsKetch (approved):
Checks out. This would've taken me ages to debug.

PR #113, opened 2026-04-08
migrate to heroicons v2

steven-klein:
# What this does

- migrates to @heroicons v2
- moves to externalize heroicons dependency
- adds import path alias for heroicons

### WTF

I've held off on this because Heroicons took a hard turn toward size optimized imports.  This technically existed in outline vs solid styles in v1 but was left to the consumer to decide on how to handle when to use each style based on the expected rendered size of the icon.

v2 now has 4 different import paths - 3 for solid: `@heroicons/vue/{24|20|16}/solid` and 1 for outline: `@heroicons/vue/24/outline`.  I don't find this optimization valuable to us and want to be able to maintain a single import path for solid vs outline icons as it works now.  This is accomplished by using some typescript declarations combined with path aliases for vite and typescript.  Basically keeping our icon imports pinned to the 24 size for solid and outline.

I like this because it means no real changes from the DX perspective other than icons having new names and it's not something I see as a maintenance issue, it's a one time configuration (the projects would also need to add config handling).

So tradeoff is some upfront configuration and icons usage looks and feels the same from a DX perspective and we get the new icons going forward.  Additionally, this alias configuration would be an easy search and replace to bail on, and it doesn't exclude the use of the native paths, if we ever did find them useful or want to experiment, like micro icons for a WYSIWYG editor or something.

Alternatively, we can just have a "rule" that you always import from the `/24/` paths - and could even consider enforcing that with a linter rule.

FWIW - the new icons do have that "fresh" look to them - so that's something.

### Project Level Configuration Looks Like:

- add heroicons.d.ts
- add vite path aliases
- add tsconfig path aliases
- (optional) have projects directly install heroicons as a dependency - still debating this one


DavidLarsKetch (approved):
> and it doesn't exclude the use of the native paths

This seems like a great approach to me gven the fuller set is still available (for specialized cases).

PR #114, opened 2026-05-14
[DEV-4554] action items updates

steven-klein:
[Ticket Details](https://www.notion.so/xylabs/8f16954aa3494675837f0cceb6a21c95?v=9d015d84f6fa48fe837c8ad0f6ab5be8&p=348d6eba0f7b80b5ab4afd3a3e4e7223&pm=s)

The core issue this change is addressing is having the ability to render anchor tag elements in `TableActions` to preserve default browser behavior a la CMD + click for new tabs etc.  I've extended the scope a little to enable this same behavior on the existing `ActionDropdown` component which also has a few instances of `window.location.href = "/url"` attached to it.

# What this does

- refactors types around ActionItems and TableActionItems - favoring duplication over complicated generics and interface extensions
- extracts the `TableActionButtons` into a generic `ActionsButtonGroup` component and handle `ActionItemLink`
- updates `ActionDropdown` to support additional features currently being used on Portal and handle `ActionItemLink`
- exports additional interfaces for consumers
- fixes issue where empty space was rendered on `DynamicTable`

See [#2717](https://github.com/xy-planning-network/college-try/pull/2717) for usage in Portal


src/composables/table.ts
```diff
@@ -37,56 +41,99 @@ export interface DynamicTableAPI<T = TableRowData> {
...
+ * @template T The shape of the underlying row data.
+ */
+export interface TableActionButton<T = TableRowData> {
+  label: string
+  attrs?: never
   disabled?: boolean | ((rowData: T, rowIndex: number) => boolean)
+  icon?: FunctionalComponent | RenderFunction
+  openInTab?: never
+  show?: boolean | ((rowData: T, rowIndex: number) => boolean)
+  url?: never
```
steven-klein, line 65: The use of `never` ensures that typescript can tell the difference between `TableActionButton` and `TableActionLink` when the type union `TableActionItem` is used.

src/composables/nav.ts
```diff
@@ -23,12 +23,39 @@ export interface Pagination {
...
   label: string
-  show?: boolean | ((...args: any[]) => boolean)
+  url: string
+  attrs?: Record<string, string | number | boolean>
+  disabled?: boolean
+  icon?: FunctionalComponent | RenderFunction
+  openInTab?: boolean
+  show?: boolean
+  onClick?: (e: Event) => void
+}
```
steven-klein, line 69: `onClick` may look unnecessary, but there are use cases for applying a click handler to an `a` tag and this will ensure that is available.  It's also common to intercept the `Event` in those cases, which is why you will see the event being passed along through out this change.

src/composables/table.ts
DavidLarsKetch: Throw that in as a code comment?

DavidLarsKetch (approved):
Looks great to me!

PR #122, opened 2026-08-05
feat: migrate to eslint@latest

steven-klein:
# What this does

- migrates to eslint@latest
- fixes ActionsDropdown transition animation (revealed by linter)
- fixes latest lint issues

### Notes

The migration was fairly straightforward in terms of migrating our previous config to the latest config.  A few rules have changed in the rulesets that we've imported, but overall the applied rules appear to be very close to what we've been working with.

I've added one new exception `"no-useless-assignment": "off",` to continue to allow us to initialize variables with their empty value.  e.g.  `let val = ""`


config/eslint.mjs
```diff
@@ -0,0 +1,36 @@
...
+      pluginVue.configs["flat/recommended"],
+      vueTsConfigs.recommended,
+    ],
+    languageOptions: {
+      ecmaVersion: "latest",
+      sourceType: "module",
+      globals: globals.browser,
+    },
+    rules: {
+      "no-useless-assignment": "off",
```
profsmallpine, line 24: 👌

profsmallpine (approved):
🎆

